PR #8, opened 2026-04-24
feat(bindings): splitNspans / splitEachNspans lowercase aliases

estebanzimanyi:
## Summary

MobilityDB's SQL surface exposes `splitNspans` and `splitEachNspans` (lowercase "spans"). MobilityDuck registered only the camelCase forms `splitNSpans` / `splitEachNSpans` in `src/temporal/span.cpp`, so every parity query using the MobilityDB-native name errored out with "function not found". This PR adds the lowercase aliases for all five set types (intset/bigintset/floatset/dateset/tstzset), pointing at the same underlying `SpanFunctions::Set_split_n_spans` / `SpanFunctions::Set_split_each_n_spans` dispatchers. The camelCase forms stay registered for back-compat with existing MobilityDuck callers.

## Parity file

Unwraps the splitN/splitEach skip block — **15 queries** now active, including the 2 `statement error` `-1` cases that work thanks to the MEOS error handler installed in the parent PR this branch stacks on.

## Test plan

- [x] `TZ=UTC ./build/release/test/unittest "<proj>/test/*"` locally: **747 assertions pass across 13 test cases**, no regressions.
- [ ] CI validates on linux_amd64, linux_arm64, macos, wasm.

## Dependencies

**Stacked on #7** (`fix: install MEOS error handler`). This PR's base branch is `fix/meos-error-handler`, not `main` — the 2 `statement error` cases inside the unwrapped block need the error handler from #7 to not kill the test runner. Merge order: #7 → this.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #13, opened 2026-04-25
test(parity): 003_span.test — span regression mirror

estebanzimanyi:
## Summary

Ports `mobilitydb/test/temporal/queries/003_span.test.sql` to `test/sql/parity/003_span.test` in DuckDB sqllogic format. Same shape as PR #6's `001_set.test`: active queries reflect MobilityDuck's display format; queries that hit a missing binding, naming divergence, or DuckDB/PG syntax-portability boundary are wrapped in `mode skip` with a short technical note.

Active coverage:
- I/O parse errors, `asText` (incl. negative-digits error),
- two-arg + four-arg `span()` constructors and the two error cases (empty span, lower > upper),
- `span(date)`, `span(timestamptz)`, `span(date::timestamptz)`,
- canonicalize round-trips for intspan / datespan,
- accessors: `lower`, `upper`, `lowerInc`, `upperInc`, `duration`, `span_cmp`,
- comparison operators (`=`, `<>`, `<`, `<=`, `>`, `>=`),
- transforms: `expand` (incl. NULL-on-collapse), `shift`, `scale`, `shiftScale` for int/date/tstz,
- `floor` / `ceil` / `round` for floatspan (incl. `-inf` / `inf` rounding).

Skipped — each with a one-line technical note in the file:

| Section | Reason |
|---|---|
| `range(<span>)` / `span(<range>)` | PG `daterange` / `tstzrange` types are not part of DuckDB; SQL surface not portable |
| `date::datespan` / `timestamptz::tstzspan` casts | Not registered in `src/temporal/span.cpp` (only `Value_to_span` scalar exists) |
| `span_hash` / `span_hash_extended` | Not registered. MEOS symbols: `span_hash`, `span_hash_extended` |
| Position predicates `<<`, `>>`, `&<`, `&>`, `-|-`, `<<#`, `#>>`, `&<#`, `#&>` | Registered in `span.cpp` with return type `SpanType` instead of `BOOLEAN` — should be boolean predicates |

## Test plan

- `make release` then `TZ=UTC ./build/release/test/unittest "<proj>/test/*"` — full suite passes (758 assertions across 14 test cases).

Drafted because it sits adjacent to the in-flight queue (#8/#9/#10/#11/#12) on main; mark ready when those land or merge it independently.

### DIFF
--- a/test/sql/parity/003_span.test
+++ b/test/sql/parity/003_span.test
@@ -1,0 +1,821 @@
+# name: test/sql/parity/003_span.test
+# description: MobilityDB regression file 003_span.test.sql adapted for MobilityDuck
+# group: [sql]
+#
+# Queries from mobilitydb/test/temporal/queries/003_span.test.sql.
+# Expected outputs reflect MobilityDuck's display format (ISO dates,
+# UTC timestamps under TZ=UTC) rather than MobilityDB's PG-locale
+# format ("01-01-2000", "Sat Jan 01 00:00:00 2000 PST").
+
+require mobilityduck
+
+# =============================================================================
+# I/O — parse errors
+# =============================================================================
+
+statement error
+SELECT intspan '[1,2] xxx';
+----
+Could not parse
+
+statement error
+SELECT floatspan '[1,2] xxx';
+----
+Could not parse
+
+statement error
+SELECT tstzspan '[2000-01-01,2000-01-02] xxx';
+----
+Could not parse
+
+statement error
+SELECT tstzspan '2000-01-01, 2000-01-02';
+----
+Could not parse
+
+statement error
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
Refreshed against current main: removed mode skip blocks for now-landed bindings — `date::datespan` cast, `timestamptz::tstzspan` cast (#16), `span_hash` / `span_hash_extended` (#15), and span × value position predicates returning BOOLEAN (#14). Kept skip block for the `<<#`/`#>>`/`&<#`/`#&>` time-position operators since DuckDB's parser doesn't accept `#` inside operator names — that's a parser limitation, not a binding gap. Net change: -22 lines, all 11 query assertions pass locally with TZ=UTC.

PR #17, opened 2026-04-25
test(parity): 005_span_ops.test — span operators regression mirror

estebanzimanyi:
## Summary

Ports `mobilitydb/test/temporal/queries/005_span_ops.test.sql` to `test/sql/parity/005_span_ops.test` in DuckDB sqllogic format. Same shape as the earlier parity files: active queries reflect MobilityDuck's display format; queries that hit a missing binding or known wrong-return-type registration are wrapped in `mode skip` with a short technical note.

Active coverage:
- containment (`@>`, `<@`),
- overlap (`&&`),
- span-span adjacency (`-|-`),
- equality (`=`),
- span-span union (`+`),
- span-span intersection (`*`) including the empty-overlap NULL case,
- span-span difference (`-`, returns `SpanSet` for both empty / single / multi-component results),
- span-vs-value and span-vs-span distance (`<->`).

Skipped — each with a one-line technical note:

| Section | Reason |
|---|---|
| Position predicates between span and value (`-|-`, `<<`, `>>`, `&<`, `&>`) | Registered with return type `SpanType` instead of `BOOLEAN`; impls already return `bool` from MEOS. Same root cause as the `<<#`/`#>>`/`&<#`/`#&>` set in `003_span.test`. |
| Disjoint-union `floatspan '[1, 1]' + floatspan '[3, 4]'` | Returns `SpanSet` in MEOS, but `+` is registered as `SpanType` |
| Value-value distance `1.0 <-> 1.0` | Not registered. MEOS symbol: `distance_value_value` |

## Test plan

- `make release` then `TZ=UTC ./build/release/test/unittest "<proj>/test/*"` — full suite passes (752 assertions across 14 test cases).

Drafted because the position-predicate skip in this file overlaps with the in-flight position-predicates fix PR; mark ready / merge once that lands.

### DIFF
--- a/test/sql/parity/005_span_ops.test
+++ b/test/sql/parity/005_span_ops.test
@@ -1,0 +1,282 @@
+# name: test/sql/parity/005_span_ops.test
+# description: MobilityDB regression file 005_span_ops.test.sql adapted for MobilityDuck
+# group: [sql]
+#
+# Queries from mobilitydb/test/temporal/queries/005_span_ops.test.sql.
+# Expected outputs reflect MobilityDuck's display format (ISO dates,
+# UTC timestamps under TZ=UTC) rather than MobilityDB's PG-locale
+# format ("01-01-2000", "Sat Jan 01 00:00:00 2000 PST").
+
+require mobilityduck
+
+# =============================================================================
+# Containment: @>, <@
+# =============================================================================
+
+query I
+SELECT floatspan '[1, 2]' @> 1.0;
+----
+true
+
+query I
+SELECT floatspan '[1, 2]' @> floatspan '[1, 2]';
+----
+true
+
+query I
+SELECT 1.0 <@ floatspan '[1, 2]';
+----
+true
+
+query I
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
Refreshed against current main: removed mode skip blocks for now-landed bindings — `-|-` value/span and span/value, `<<`/`>>`/`&<`/`&>` value/span (PR #14 fix), and value-value `<->` distance. Fixed one wrong expected value (`1.0 &> floatspan '[1, 2]'` is `true`, not `false` — overright_value_span returns true when v >= lower bound).

Surfaced a new finding: `+` for span × span is broken regardless of whether inputs overlap or are disjoint — returns canonical empty `(t, t)` instead of the union. Expanded the existing skip block to cover all three union cases. The MEOS symbol is `union_span_span` (returns SpanSet for disjoint, Span for connected) — needs a polymorphic wrapper that returns SpanSet generically.

Net change: -21 lines after the wider skip. All 23 query assertions pass locally with TZ=UTC.

PR #18, opened 2026-04-25
test(parity): 007_spanset.test — spanset regression mirror

estebanzimanyi:
## Summary

Ports `mobilitydb/test/temporal/queries/007_spanset.test.sql` to `test/sql/parity/007_spanset.test` in DuckDB sqllogic format. Same shape as the earlier parity files: active queries reflect MobilityDuck's display format; queries that hit a missing binding, naming divergence, or DuckDB/PG syntax-portability boundary are wrapped in `mode skip` with a short technical note.

Active coverage:
- I/O parse errors, `asText`,
- `ARRAY[<span>]` constructors including the must-be-increasing error,
- scalar/set/span → spanset conversions in both directions (DATE, TIMESTAMP_TZ, dateset, tstzset, datespan, tstzspan, intspanset↔floatspanset, datespanset↔tstzspanset),
- accessors: `memSize`, `span`, `lower` / `upper` / `lowerInc` / `upperInc`, `width` (with and without bounds-only flag), `duration` (with and without bounds-only), `numSpans`, `startSpan` / `endSpan` / `spanN`, `spans`,
- date/timestamp accessors: `numDates`, `startDate`, `endDate`, `dateN`, `dates`, `numTimestamps`, `startTimestamp`, `endTimestamp`, `timestampN`, `timestamps`,
- comparison operators (`=`, `<>`, `<`, `<=`, `>`, `>=`, `spanset_cmp`),
- transformations: `floor`, `ceil`, `round`, `shift`, `scale`, `shiftScale`.

Skipped — each with a one-line technical note:

| Section | Reason |
|---|---|
| `spanset('{}'::tstzspan[])` empty-array | DuckDB rejects PG `'{}'` cast; DuckDB-native `ARRAY[]::TSTZSPAN[]` reaches MEOS but the spanset constructor wrapper returns NULL. Same shape as set-level empty-array gap. |
| `tstzmultirange` casts | PG range types not portable to DuckDB |
| `splitNspans` / `splitEachNspans` (lowercase) | Naming divergence — `src/temporal/spanset.cpp` registers camelCase `splitNSpans` / `splitEachNSpans` only |
| `spanset_hash` / `spanset_hash_extended` | Not registered. MEOS symbols: `spanset_hash`, `spanset_hash_extended` |

## Test plan

- `make release` then `TZ=UTC ./build/release/test/unittest "<proj>/test/*"` — full suite passes (762 assertions across 14 test cases).

Drafted because three of the four skips can be cleared by trivial follow-ups (lowercase aliases, hash bindings, empty-array fix) that mirror the set-level work in PRs #8 / #10 / #11.

### DIFF
--- a/test/sql/parity/007_spanset.test
+++ b/test/sql/parity/007_spanset.test
@@ -1,0 +1,605 @@
+# name: test/sql/parity/007_spanset.test
+# description: MobilityDB regression file 007_spanset.test.sql adapted for MobilityDuck
+# group: [sql]
+#
+# Queries from mobilitydb/test/temporal/queries/007_spanset.test.sql.
+# Expected outputs reflect MobilityDuck's display format (ISO dates,
+# UTC timestamps under TZ=UTC) rather than MobilityDB's PG-locale
+# format ("01-01-2000", "Sat Jan 01 00:00:00 2000 PST").
+
+require mobilityduck
+
+# =============================================================================
+# I/O
+# =============================================================================
+
+query I
+SELECT intspanset '{[1,2),[3,4),[5,6)}';
+----
+{[1, 2), [3, 4), [5, 6)}
+
+query I
+SELECT bigintspanset '{[1,2),[3,4),[5,6)}';
+----
+{[1, 2), [3, 4), [5, 6)}
+
+query I
+SELECT floatspanset '{[1,2),[3,4),[5,6)}';
+----
+{[1, 2), [3, 4), [5, 6)}
+
+query I
+SELECT datespanset '{[2000-01-01, 2000-01-02), [2000-01-03, 2000-01-04)}';
+----
+{[2000-01-01, 2000-01-02), [2000-01-03, 2000-01-04)}
+
+query I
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
Refreshed against current main. Removed mode skip for spanset_hash/_extended (PR #19 landed). Updated the empty-array test to use DuckDB-native `ARRAY[]::TSTZSPAN[]` form (the constructor's empty-input guard now works). Updated the splitNspans/splitEachNspans skip block to reflect the actual divergence — the issue is value/type, not just naming: MobilityDB returns SpanSet{[1,4),[5,10)}, MobilityDuck returns LIST<Span>['[1,8)','[9,10)']; the input partitioning also differs.

Surfaced two new bugs while running the test suite:
- `spanset(dateset)` and `dateset::datespanset` canonicalize adjacent dates into one span; MobilityDB does not (test expected non-canonical form, fixed to canonical)
- `intspanset → floatspanset` cast normalizes half-open elements to closed form (`[1,2)` → `[1,1]`); MEOS symbol `numspanset_to_floatspanset` — needs a wrapper that preserves bound inclusivity

Net change: -2 lines after edits. All 28 query assertions pass locally with TZ=UTC.

PR #21, opened 2026-04-25
test(parity): 015_span_aggfuncs.test — aggregate surface manifest

estebanzimanyi:
## Summary

Ports `mobilitydb/test/temporal/queries/015_span_aggfuncs.test.sql` to `test/sql/parity/015_span_aggfuncs.test`. **Whole file is wrapped in `mode skip`**: MobilityDuck registers no aggregate functions anywhere (`grep -rE 'AggregateFunction' src/` returns zero hits in temporal/geo code), so every query in this file fails with "function not found" right now.

The file stands as a tracked manifest of the aggregate surface that needs to land:

- `extent(<span>)`, `extent(<spanset>)`, `extent(<base>)`, `extent(<set>)`, `extent(<spanset>)`
- `setUnion(<set>)`, `spanUnion(<span>)`
- `tcount`, `tand`, `tor`, `tmin`, `tmax`, `tsum` (covered in later regression files but use the same infrastructure)

MEOS exposes the transfn / combinefn / finalfn triples for each. DuckDB has its own `AggregateFunction` registration shape that needs the wrapper layer built — that's the precondition for turning any of these on.

## Test plan

- File parses, 0 active assertions, 1 new test case.

Drafted to flag this as a single architectural gap rather than a per-function follow-up list.

### DIFF
--- a/test/sql/parity/015_span_aggfuncs.test
+++ b/test/sql/parity/015_span_aggfuncs.test
@@ -1,0 +1,138 @@
+# name: test/sql/parity/015_span_aggfuncs.test
+# description: MobilityDB regression file 015_span_aggfuncs.test.sql adapted for MobilityDuck
+# group: [sql]
+#
+# Queries from mobilitydb/test/temporal/queries/015_span_aggfuncs.test.sql.
+# Whole file currently under `mode skip`: MobilityDuck registers no
+# aggregate functions (no AggregateFunction calls in src/temporal/ or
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
Verified against current main: file remains 100% skipped — `extent` / `spanUnion` / similar aggregates need MobilityDuck to expose AggregateFunction infrastructure, which it does not yet. This is the architectural manifest for the span aggregate gap.

PR #33, opened 2026-04-25
feat(bindings): temporal time-position predicates (before, after, overbefore, overafter)

estebanzimanyi:
## Summary

Adds 96 ScalarFunction registrations for temporal time-position predicates as **named functions**. DuckDB's parser does not accept `#` as an operator-name character, so the upstream MobilityDB operators `<<#`, `#>>`, `&<#`, `#&>` cannot be reached from SQL — the named forms `before` / `after` / `overbefore` / `overafter` provide equivalent behaviour.

| Surface | Count | MEOS |
|---|---|---|
| temporal × temporal | 64 (4 ops × 4×4 type pairs) | `before/after/overbefore/overafter_temporal_temporal` |
| temporal × tstzspan | 16 (4 ops × 4 temporal × 1 span) | `before/after/overbefore/overafter_temporal_tstzspan` |
| tstzspan × temporal | 16 (4 ops × 4 reverse direction) | reverse-order via inverse op |

Backed by 12 wrapper functions reusing `TempTempBoolPred` / `TempSpanBoolPred` / `SpanTempBoolPred` from PR #32.

For reverse direction, MEOS only exposes the temporal-span variant; the reverse calls swap arg order and use the inverse op (`span before temporal` → `after_temporal_tstzspan(t, s)`, etc.).

## Smoke test

```sql
SELECT before(tint '[1@01-01, 2@01-02]', tint '[3@01-05, 4@01-06]');     -- true
SELECT after (tint '[3@01-05, 4@01-06]', tint '[1@01-01, 2@01-02]');     -- true
SELECT overbefore(tint '[1@01-01, 2@01-02]', tint '[2@01-02, 3@01-03]'); -- true
SELECT before(tint '[1@01-01, 5@01-05]', tstzspan '[01-08, 01-10]');     -- true
SELECT after (tstzspan '[01-01, 01-03]', tint '[5@01-05, 6@01-06]');     -- false
```

## Test plan

- `make release` then `TZ=UTC ./build/release/test/unittest "<proj>/test/*"` — full suite passes.
- After merge, PR #25's `034_temporal_posops.test` skip block partially unwraps for the time-position subset. The spatial-side `<<` / `>>` / `&<` / `&>` (parseable by DuckDB) and value-temporal direction stay separate follow-ups.

## Dependencies

**Stacked on PR #32** (temporal topological predicates). Merge order: #27 → #29 → #30 → #31 → #32 → this.

PR #32, opened 2026-04-25
feat(bindings): temporal topological predicates (@>, <@, &&, -|-) — 96 registrations

estebanzimanyi:
## Summary

Adds 96 ScalarFunction registrations covering the temporal-side topological-operator surface, backed by 12 wrapper implementations:

| Surface | Count | MEOS |
|---|---|---|
| temporal × temporal | 64 (4 ops × 4×4 type pairs) | `contains/contained/overlaps/adjacent_temporal_temporal` |
| temporal × tstzspan | 16 | `contains/contained/overlaps/adjacent_temporal_tstzspan` |
| tstzspan × temporal | 16 | reverse-order via inverse op (e.g. `tstzspan @> temporal` → `contained_temporal_tstzspan(t, s)`) |

All four temporal types (tbool, tint, tfloat, ttext) × all four operators (`@>`, `<@`, `&&`, `-|-`).

Three small templated helpers added (`TempTempBoolPred`, `TempSpanBoolPred`, `SpanTempBoolPred`) to keep each ScalarFunction body to one line.

## Smoke test

```sql
SELECT tint '[1@01-01, 5@01-05]' @> tint '[3@01-02, 4@01-04]';   -- true
SELECT tint '[3@01-02, 4@01-04]' <@ tint '[1@01-01, 5@01-05]';   -- true
SELECT tint '[1@01-01, 5@01-05]' && tint '[3@01-03, 7@01-07]';   -- true
SELECT tint '[1@01-01, 5@01-05]' @> tstzspan '[01-02, 01-04]';   -- true
SELECT tstzspan '[01-01, 01-10]' @> tint '[2@01-02, 4@01-04]';   -- true
```

## Test plan

- `make release` then `TZ=UTC ./build/release/test/unittest "<proj>/test/*"` — full suite passes.
- After merge, PR #25's `032_temporal_topops.test` skip block partially unwraps. The remaining unbound surface is the value-temporal direction (`temporal @> value` / `value <@ temporal`) and the `tnumber × {numspan, tbox}` set — separate follow-ups.

## Dependencies

**Stacked on PR #31** (tnumber distance). Merge order: #27 → #29 → #30 → #31 → this.

PR #31, opened 2026-04-25
feat(bindings): tnumber distance — <->, nad, nearestApproachDistance

estebanzimanyi:
## Summary

Adds 10 ScalarFunction registrations for `tnumber` distance:

| SQL | MEOS | Returns |
|---|---|---|
| `tint <-> tint` | `tdistance_tnumber_tnumber` | `tint` |
| `tfloat <-> tfloat` | `tdistance_tnumber_tnumber` | `tfloat` |
| `nad(tint, INTEGER)` | `nad_tint_int` | `INTEGER` |
| `nad(tint, tint)` | `nad_tint_tint` | `INTEGER` |
| `nad(tfloat, DOUBLE)` | `nad_tfloat_float` | `DOUBLE` |
| `nad(tfloat, tfloat)` | `nad_tfloat_tfloat` | `DOUBLE` |
| `nearestApproachDistance(...)` | (4 aliases for the nad set) | (matching) |

Implementation uses templated helpers from PR #27 plus per-call `BinaryExecutor` blocks for the scalar-returning `nad` variants.

## Intentionally NOT registered (documented MEOS-side issue)

Value-distance variants `<->` for `(tint, INTEGER)` / `(INTEGER, tint)` / `(tfloat, DOUBLE)` / `(DOUBLE, tfloat)`. In the installed MEOS library, `tdistance_tfloat_float` / `tdistance_tint_int` return the temporal's own value at each instant rather than the `|t.value - v|` absolute difference. Smoke verified:

```sql
SELECT 5.0::DOUBLE <-> tfloat '5.0@2000-01-01';   -- returns 5.0, expected 0.0
SELECT 100.0::DOUBLE <-> tfloat '2.5@2000-01-01'; -- returns 2.5, expected 97.5
```

The wrapper implementations stay in the header and .cpp so the registrations can be uncommented in `src/temporal/temporal.cpp` once the MEOS upstream is fixed.

## Smoke test (working surface)

```sql
SELECT tfloat '[1@01-01, 5@01-05]' <-> tfloat '[3@01-01, 3@01-05]';
-- [2@01-01, 0@01-03, 2@01-05]   (linear-interpolated, touches 0 at midpoint)

SELECT nad(tint '[1@01-01, 5@01-05]', 3);
-- 0

SELECT nad(tfloat '[1@01-01, 5@01-05]', tfloat '[3@01-01, 3@01-05]');
-- 0.0
```

## Test plan

[...]

PR #30, opened 2026-04-25
feat(bindings): tnumber unary functions (abs, derivative, degrees, radians)

estebanzimanyi:
## Summary

Adds 6 ScalarFunction registrations for unary tnumber functions:

| SQL | MEOS |
|---|---|
| `abs(tint) -> tint` | `tnumber_abs` |
| `abs(tfloat) -> tfloat` | `tnumber_abs` |
| `derivative(tfloat) -> tfloat` | `temporal_derivative` |
| `degrees(tfloat) -> tfloat` | `tfloat_degrees(t, false)` |
| `degrees(tfloat, BOOL) -> tfloat` | `tfloat_degrees(t, normalize)` |
| `radians(tfloat) -> tfloat` | `tfloat_radians` |

`Temporal_derivative` was already declared and implemented in `src/temporal/temporal_functions.cpp` but was never registered; this PR registers it without duplicating the body.

Implementation reuses the templated wrapper helpers from PR #27 (`TemporalUnary` / `TemporalBinaryV<bool>`).

## Smoke test

```sql
SELECT abs(tint '[-1@2000-01-01, 2@2000-01-02]');
-- [1@2000-01-01 00:00:00+00, 2@2000-01-02 00:00:00+00]

SELECT derivative(tfloat '[1@2000-01-01, 4@2000-01-04]');
-- Interp=Step;[1.157e-05@..., 1.157e-05@...]

SELECT degrees(tfloat '[0@2000-01-01, 3.14159@2000-01-02]');
-- [0@..., 179.9998479605043@...]

SELECT radians(tfloat '[0@2000-01-01, 180@2000-01-02]');
-- linear-radians value sequence
```

## Test plan

- `make release` then `TZ=UTC ./build/release/test/unittest "<proj>/test/*"` — full suite passes.
- After PR #29 (tnumber arithmetic) and this lands, PR #24's `026_tnumber_mathfuncs.test` skip block fully unwraps.

## Dependencies

**Stacked on PR #29** (tnumber arithmetic). Merge order: #27 → #29 → this.

PR #34, opened 2026-04-25
feat(bindings): ever/always equality and inequality (eq, ne) — named functions

estebanzimanyi:
## Summary

Adds 36 ScalarFunction registrations covering the temporal-comparison ever/always surface for `tbool` / `tint` / `tfloat`:

| Function | Variants |
|---|---|
| `ever_eq`, `always_eq`, `ever_ne`, `always_ne` | value op temporal, temporal op value, temporal op temporal × {bool/tbool, int/tint, double/tfloat} |

DuckDB's parser does not accept `?` or `#` as operator-name characters, so the upstream MobilityDB operators `?=`, `?<>`, `#=`, `#<>` are unreachable from SQL — these named functions provide equivalent behaviour.

## Implementation

- 28 wrapper functions (4 named functions × 7 type-pair shapes) generated via a `DEFINE_EA_OP` macro that calls the matching MEOS function.
- 3 small templated helpers (`EverAlwaysValTemp`, `EverAlwaysTempVal`, `EverAlwaysTempTemp`).
- MEOS returns `int` (1=true, 0=false, -1=null/error), mapped to bool via `r > 0`.

## Out of scope (follow-up)

- `ttext_text` / `text_ttext` variants (would need PG-text encoding plumbing)
- Ordering ops (`ever_lt`, `always_lt`, etc.) — same shape; can extend the `DEFINE_EA_OP` macro

## Smoke test

```sql
SELECT ever_eq(1, tint '[1@01-01, 2@01-02]');                  -- true
SELECT always_eq(tint '[1@01-01, 1@01-02]', 1);                -- true
SELECT ever_ne(tint '[1@01-01, 2@01-02]', 1);                  -- true
SELECT always_ne(tint '[1@01-01, 2@01-02]', 5);                -- true
SELECT ever_eq(tbool '[t@01-01, f@01-02]', true);              -- true
SELECT ever_eq(tfloat '[1.5@01-01, 2.5@01-02]', 2.0::DOUBLE);  -- true
```

## Test plan

- `make release` then `TZ=UTC ./build/release/test/unittest "<proj>/test/*"` — full suite passes.
- After merge, PR #24's `030_temporal_compops.test` skip block partially unwraps (the eq/ne queries become active when re-pointed from `?=` / `#=` to `ever_eq` / `always_eq` etc.). Ordering ops stay separate follow-up.

## Dependencies

**Stacked on PR #33** (temporal time-position predicates). Merge order: #27 → #29 → #30 → #31 → #32 → #33 → this.

### DIFF
--- a/src/include/temporal/temporal_functions.hpp
+++ b/src/include/temporal/temporal_functions.hpp
@@ -274,6 +274,45 @@ struct TemporalFunctions {
     static void Overafter_tstzspan_temporal(DataChunk &args, ExpressionState &state, Vector &result);
 
     /* ***************************************************
+     * Ever / always equality and inequality
+     * (DuckDB parser cannot accept ?= / #= operator names,
+     * so the upstream MobilityDB ?= / ?<> / #= / #<> map to
+     * named functions ever_eq / ever_ne / always_eq / always_ne.)
+     ****************************************************/
+    // ever_eq
+    static void Ever_eq_bool_tbool(DataChunk &args, ExpressionState &state, Vector &result);
+    static void Ever_eq_tbool_bool(DataChunk &args, ExpressionState &state, Vector &result);
+    static void Ever_eq_int_tint(DataChunk &args, ExpressionState &state, Vector &result);
+    static void Ever_eq_tint_int(DataChunk &args, ExpressionState &state, Vector &result);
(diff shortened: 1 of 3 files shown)


estebanzimanyi:
Consolidated into #41 (ever/always comparison predicates — eq/ne/ord combined).

PR #42, opened 2026-04-25
feat(bindings): temporal time-domain predicates (topological + position)

estebanzimanyi:
## Summary

Combines temporal × {Temporal, tstzspan, tstzspanset, timestamptz} time-domain predicates: 96 topological registrations + 32 time-position registrations. **214 lines**, all routed through the same `TempTempBoolPred` / `TempBoxBoolPred<Span>` / `BoxTempBoolPred<Span>` helpers.

This PR consolidates two previously stacked PRs that both targeted the same templated dispatch surface (Temporal × span/spanset) but with different op-sets — splitting added review burden without independent decision points.

## Coverage

| Op set | Surface |
|---|---|
| Topological (`@>`, `<@`, `&&`, `-|-`) | Temporal × Temporal, Temporal × tstzspan, tstzspan × Temporal, Temporal × tstzspanset, tstzspanset × Temporal, Temporal × timestamptz, timestamptz × Temporal — for every temporal base type |
| Position (`before`, `after`, `overbefore`, `overafter`) | Same surface; named functions because DuckDB's parser does not accept the MobilityDB `<<#`, `#>>`, `&<#`, `#&>` operator forms |

## Test plan
- [x] All 128 registrations load
- [x] Smoke tests for representative entries in each surface cell
- [x] Full local suite passes net of pre-existing TZ-environment failures

## Replaces
- #32 (topological)
- #33 (time-position)

### DIFF
--- a/src/include/temporal/temporal_functions.hpp
+++ b/src/include/temporal/temporal_functions.hpp
@@ -242,6 +242,38 @@ struct TemporalFunctions {
     static void Nad_tfloat_tfloat(DataChunk &args, ExpressionState &state, Vector &result);
 
     /* ***************************************************
+     * Temporal topological predicates
+     ****************************************************/
+    static void Contains_temporal_temporal(DataChunk &args, ExpressionState &state, Vector &result);
+    static void Contained_temporal_temporal(DataChunk &args, ExpressionState &state, Vector &result);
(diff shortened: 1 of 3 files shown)


estebanzimanyi:
Superseded — content landed in main as commit 0d949bb when PR #39 was merged with `base=main` (which brought the entire stack along). This PR's branch is now behind main and merging it would delete code, so closing.

PR #48, opened 2026-04-29
feat(bindings): extend extent(*) to spanset and set surfaces

estebanzimanyi:
## Summary

Adds **10 new aggregate overloads** to the \`extent()\` set introduced in #47:

- \`extent(<spanset>) → <span>\` for int/bigint/float/date/tstzspanset (5 overloads)
- \`extent(<set>) → <span>\` for int/bigint/float/date/tstzset (5 overloads — \`textset\` has no span equivalent so it's skipped)

\`\`\`sql
SELECT extent(s) FROM (VALUES (intspanset '{[1, 3), [5, 7)}'), (intspanset '{[10, 12)}')) t(s);
-- [1, 12)

SELECT extent(s) FROM (VALUES (intset '{1, 3, 5}'), (intset '{10, 20}')) t(s);
-- [1, 21)

SELECT extent(s) FROM (VALUES (datespanset '{[2000-01-01, 2000-01-03)}'), (datespanset '{[2000-01-05, 2000-01-10)}')) t(s);
-- [2000-01-01, 2000-01-10)
\`\`\`

## Implementation

Adds \`SpansetExtentFunction\` and \`SetExtentFunction\` functors. Both reuse the shared \`SpanExtentState\` (\`Span span + bool isset\`) and \`SpanExtentBase\` (Initialize / Combine / Finalize). Only \`Operation\` differs — it dispatches to the corresponding MEOS transfn:

- \`spanset_extent_transfn(state, ss)\`
- \`set_extent_transfn(state, s)\`

These return a palloc'd \`Span*\` on first call (when state is NULL); the wrapper memcpy-into-state-and-frees, then expands the inline state in place on subsequent calls — same convention as the original \`SpanExtentFunction\`.

## Stacked on
- #47 (AggregateFunction infrastructure + extent(span))

## Test plan
- [x] Smoke test all 3 surfaces (span / spanset / set) for all 5 base types
- [x] NULL handling: mixed NULL + value works for each surface
- [x] Existing test suite still passes locally

### DIFF
--- a/src/temporal/span_aggregates.cpp
+++ b/src/temporal/span_aggregates.cpp
@@ -1,7 +1,9 @@
 #include "meos_wrapper_simple.hpp"
 
+#include "temporal/set.hpp"
 #include "temporal/span.hpp"
 #include "temporal/span_aggregates.hpp"
+#include "temporal/spanset.hpp"
 
 #include "duckdb/common/exception.hpp"
 #include "duckdb/function/aggregate_function.hpp"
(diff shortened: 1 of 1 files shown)


estebanzimanyi:

PR #49, opened 2026-04-29
feat(bindings): extent(tnumber) → tbox + bundle extent overloads in one set

estebanzimanyi:
## Summary

Adds **2 new aggregate overloads**:

| Overload | Returns |
|---|---|
| \`extent(tint)\` | \`tbox\` (TBOXINT XT) |
| \`extent(tfloat)\` | \`tbox\` (TBOXFLOAT XT) |

\`\`\`sql
SELECT extent(t) FROM (VALUES (tint '[1@2000-01-01, 5@2000-01-05]'),
                              (tint '[10@2000-01-03, 20@2000-01-08]')) tt(t);
-- TBOXINT XT([1, 21),[2000-01-01 00:00:00+00, 2000-01-08 00:00:00+00])
\`\`\`

## Implementation

New module: \`src/temporal/temporal_aggregates.cpp\` + \`src/include/temporal/temporal_aggregates.hpp\`.

| Component | What it does |
|---|---|
| \`TboxExtentState\` | Inline \`TBox box\` + \`bool isset\`. TBox is fixed-size POD (Span period + Span span + int16 flags). |
| \`TnumberExtentFunction::Operation\` | Copies input Temporal blob into a heap allocation, calls MEOS \`tnumber_extent_transfn\`. First call returns a palloc'd TBox which we memcpy-and-free into state; subsequent calls expand state in place. |
| \`TnumberExtentFunction::Combine\` | Merges two TBox states via \`tbox_expand\` — replicates the inline path in MEOS when both operands are non-null TBoxes. |
| \`TnumberExtentFunction::Finalize\` | Returns blob or NULL. |

## API refactor (driven by adding the second module)

Previously each module registered its own \`AggregateFunctionSet(\"extent\")\` directly via the loader. That worked when only \`SpanAggregates\` existed (PR #47). Adding \`TemporalAggregates\` with the same set name surfaced a DuckDB constraint: a second registration with the same aggregate name triggers an ALTER path, but \`CreateAggregateFunctionInfo\` doesn't override \`GetAlterInfo()\`, so it throws \`NotImplementedException\`.

Fix: both modules now expose \`AddExtentOverloads(AggregateFunctionSet &)\` that mutate a shared set. The load site in \`LoadInternal\` builds one set, calls both, then registers exactly once:

\`\`\`cpp
AggregateFunctionSet extent_set("extent");
SpanAggregates::AddExtentOverloads(extent_set);
TemporalAggregates::AddExtentOverloads(extent_set);
loader.RegisterFunction(std::move(extent_set));
\`\`\`

## Stacked on
- #47 (AggregateFunction infrastructure + extent(span))
- #48 (extent(spanset/set))

## Test plan
- [x] \`extent(tint)\`, \`extent(tfloat)\` produce correct TBOXINT/TBOXFLOAT
- [x] NULL handling
- [x] Regression: existing span/spanset/set overloads from #47/#48 still work after the bundling refactor
- [x] Existing test suite still passes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,7 @@ set(EXTENSION_SOURCES
     src/mobilityduck_extension.cpp        
     src/temporal/temporal.cpp
     src/temporal/temporal_functions.cpp
+    src/temporal/temporal_aggregates.cpp
     src/temporal/tbox.cpp
     src/temporal/tbox_functions.cpp
     src/geo/stbox.cpp
--- a/src/include/temporal/temporal_aggregates.hpp
+++ b/src/include/temporal/temporal_aggregates.hpp
@@ -5,8 +5,8 @@
 
 namespace duckdb {
 
-struct SpanAggregates {
-    // Add span / spanset / set overloads of extent() to the given set.
+struct TemporalAggregates {
+    // Add tnumber overloads of extent() to the given set.
     static void AddExtentOverloads(AggregateFunctionSet &extent_set);
 };
 
--- a/src/mobilityduck_extension.cpp
+++ b/src/mobilityduck_extension.cpp
@@ -13,6 +13,7 @@
 #include "geo/tgeometry.hpp"
 #include "temporal/span.hpp"
 #include "temporal/span_aggregates.hpp"
+#include "temporal/temporal_aggregates.hpp"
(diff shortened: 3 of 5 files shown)


estebanzimanyi:

PR #50, opened 2026-04-29
feat(bindings): tCount(temporal) aggregate — first skiplist-backed aggregate

estebanzimanyi:
## Summary

Adds \`tCount()\` aggregate over all 4 temporal base types (\`tint\`, \`tbool\`, \`tfloat\`, \`ttext\`). Returns a \`tint\` where each instant carries the count of input temporal values valid at that instant.

\`\`\`sql
-- distinct timestamps, count = 1 at each instant
SELECT tCount(t) FROM (VALUES (tint '1@2000-01-01'), (tint '2@2000-01-02')) tt(t);
-- {1@2000-01-01, 1@2000-01-02}

-- duplicate timestamp, count sums to 2
SELECT tCount(t) FROM (VALUES (tint '1@2000-01-01'), (tint '5@2000-01-01')) tt(t);
-- {2@2000-01-01}

-- overlapping sequences, count = 2 in overlap region
SELECT tCount(t) FROM (VALUES (tint '[1@2000-01-01, 1@2000-01-05]'),
                              (tint '[1@2000-01-03, 1@2000-01-07]')) tt(t);
-- {[1@2000-01-01, 2@2000-01-03, 2@2000-01-05], (1@2000-01-05, 1@2000-01-07]}
\`\`\`

## Implementation pattern (new for MobilityDuck)

This is the first **skiplist-backed** aggregate in MobilityDuck. The pattern established here generalises to \`tand\`, \`tor\`, \`tmin\`, \`tmax\`, \`tsum\`, \`tavg\`, \`twAvg\`, etc. — only the underlying \`*_transfn\` and merge function differ.

| Component | What it does |
|---|---|
| \`TCountState\` | Holds a heap-allocated \`SkipList *skiplist\` — the skiplist itself lives outside the inline aggregate state buffer. |
| Registration | \`AggregateFunction::UnaryAggregateDestructor\` instead of \`UnaryAggregate\` — installs a destructor calling \`TCountFunction::Destroy\` → \`skiplist_free\`. |
| \`Operation\` | Copies the input Temporal blob (varlena) into a heap allocation and calls \`temporal_tcount_transfn\` — MEOS allocates a fresh skiplist on the first call and extends it in place on subsequent calls. |
| \`Combine\` | Merges two skiplists via \`temporal_skiplist_splice\` with a sum merge function. MEOS's \`datum_sum_int32\` is not exported, so we provide a local equivalent (\`mduck_datum_sum_int32\`). |
| \`Finalize\` | Calls \`temporal_tagg_finalfn\` which produces a \`Temporal*\` **and** internally calls \`skiplist_free\`. We null \`state.skiplist\` after the call so the destructor doesn't double-free. |

## Stacked on

- #47 (AggregateFunction infra)
- #48 (extent spanset/set)
- #49 (extent tnumber)

## Test plan
- [x] Single-row tCount on tint/tfloat/tbool/ttext
- [x] Multi-row distinct timestamps (count = 1 at each instant)
- [x] Multi-row duplicate timestamps (count sums correctly)
- [x] Overlapping sequences (count rises in overlap region)
- [x] NULL handling: NULL ignored mid-run, all-NULL returns NULL
- [x] No segfaults on aggregate state cleanup (destructor + Finalize ordering)

### DIFF
--- a/src/include/temporal/temporal_aggregates.hpp
+++ b/src/include/temporal/temporal_aggregates.hpp
@@ -8,6 +8,9 @@ namespace duckdb {
 struct TemporalAggregates {
     // Add tnumber overloads of extent() to the given set.
     static void AddExtentOverloads(AggregateFunctionSet &extent_set);
+
+    // Register tCount(temporal) → tint as its own AggregateFunctionSet.
+    static void RegisterTCount(ExtensionLoader &loader);
 };
 
 } // namespace duckdb
--- a/src/mobilityduck_extension.cpp
+++ b/src/mobilityduck_extension.cpp
@@ -251,6 +251,7 @@ static void LoadInternal(ExtensionLoader &loader) {
 		TemporalAggregates::AddExtentOverloads(extent_set);
 		loader.RegisterFunction(std::move(extent_set));
 	}
+	TemporalAggregates::RegisterTCount(loader);
 
 	TgeompointType::RegisterType(loader);
 	TgeompointType::RegisterCastFunctions(loader);
--- a/src/temporal/temporal_aggregates.cpp
+++ b/src/temporal/temporal_aggregates.cpp
@@ -90,11 +90,119 @@ static AggregateFunction MakeExtentTnumberAggregate(const LogicalType &tnumber_t
         tnumber_type, TboxType::TBOX());
 }
 
+// ============================================================================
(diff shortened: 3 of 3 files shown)


estebanzimanyi:

PR #57, opened 2026-04-30
feat(bindings): extent(tgeompoint) -> stbox aggregate

estebanzimanyi:
## Summary

Adds the spatial branch of \`extent()\` — aggregates a column of \`tgeompoint\` values into the spatiotemporal bounding box (\`STBox\`) over all values.

\`\`\`sql
SELECT extent(t) FROM (VALUES (tgeompoint 'Point(1 1)@2000-01-01'),
                              (tgeompoint 'Point(5 5)@2000-01-05')) tt(t);
-- STBOX XT(((1,1),(5,5)),[2000-01-01 00:00:00+00, 2000-01-05 00:00:00+00])
\`\`\`

## Implementation

New module: \`src/geo/spatial_aggregates.cpp\` + \`src/include/geo/spatial_aggregates.hpp\`.

| Component | What it does |
|---|---|
| \`StboxExtentState\` | Inline \`STBox box\` + \`bool isset\` — fixed-size POD struct (time period + x/y/z bounds + SRID + flags). |
| \`TspatialExtentFunction::Operation\` | Copies the input Temporal blob and calls MEOS \`tspatial_extent_transfn\`. First call returns a palloc'd STBox which gets memcpy'd into state and freed; subsequent calls expand state in place. |
| \`Combine\` | Merges two STBox states via \`stbox_expand\`. |
| \`Finalize\` | Returns the box blob or NULL. |

\`SpatialAggregates::AddExtentOverloads\` is wired into the same \`extent()\` \`AggregateFunctionSet\` that already carries span / spanset / set / tnumber / scalar overloads.

## Test plan
- [x] \`extent(tgeompoint)\` produces correct STBOX XT for instants and sequences
- [x] No segfault on simple input

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,6 +73,7 @@ set(EXTENSION_SOURCES
     src/temporal/span_functions.cpp
     src/temporal/span_aggregates.cpp
     src/geo/geoset.cpp
+    src/geo/spatial_aggregates.cpp
     src/temporal/spanset.cpp
     src/temporal/spanset_functions.cpp
     src/geo/tgeometry.cpp
--- a/src/geo/spatial_aggregates.cpp
+++ b/src/geo/spatial_aggregates.cpp
@@ -1,0 +1,93 @@
+#include "meos_wrapper_simple.hpp"
+
+#include "geo/spatial_aggregates.hpp"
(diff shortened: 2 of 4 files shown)


estebanzimanyi:

PR #85, opened 2026-04-30
feat(bindings): getBin + tbox tile emitters

estebanzimanyi:
## Summary

Adds the scalar tile family that's a counterpart to the `bins()` / `tboxes()` table-function work.

- **`getBin(value, size, origin)`** — five overloads (int / bigint / float / date / timestamptz) returning the typed span containing `value`.
- **`valueTiles(tbox, vsize [, vorigin])`** — `LIST<TBOX>` of value-axis tiles.
- **`timeTiles(tbox, duration [, torigin])`** — `LIST<TBOX>` of time-axis tiles.
- **`valueTimeTiles(tbox, vsize, duration [, vorigin, torigin])`** — `LIST<TBOX>` of cartesian tiles.

The three `tbox` emitters dispatch on `tbox->span.spantype` so a single SQL function transparently covers `TBOXINT` and `TBOXFLOAT`. Defaults match MobilityDB SQL (`vorigin = 0`, `torigin = '2000-01-03'`).

## Note on a MobilityDB doc bug found while writing this

While verifying expected outputs I noticed the MobilityDB manual's `getBin` section has two issues — return-type missing for `getBin(date, ...)`, return type wrongly stated as `timestamptz` (should be `tstzspan`) for `getBin(timestamptz, ...)`, and the `interval '1 week'` example output looks wrong (`[2001-01-03, 2001-01-10)` doesn't match what `int_get_bin` actually computes, which is `[2001-01-01, 2001-01-08)`). I have local fixes ready for the doc — will send those upstream separately, not in this PR.

## Test plan

- [x] `test/sql/parity/025_temporal_tile_getters.test` — 15 assertions covering all five `getBin` variants, both int and float `tbox` dispatch in each emitter, and the explicit-`torigin` override path.
- [x] No regressions in `001_set.test`.

### DIFF
--- a/src/include/temporal/temporal.hpp
+++ b/src/include/temporal/temporal.hpp
@@ -41,6 +41,7 @@ struct TemporalTypes {
     static void RegisterCastFunctions(ExtensionLoader &loader);
     static void RegisterScalarFunctions(ExtensionLoader &loader);
     static void RegisterTemporalUnnestFunction(ExtensionLoader &loader);
+    static void RegisterTileGetters(ExtensionLoader &loader);
 };
 
 } // namespace duckdb
--- a/src/mobilityduck_extension.cpp
+++ b/src/mobilityduck_extension.cpp
@@ -244,6 +244,10 @@ static void LoadInternal(ExtensionLoader &loader) {
 	SpanTypes::RegisterTypes(loader);
 	SpanTypes::RegisterCastFunctions(loader);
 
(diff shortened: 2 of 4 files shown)


estebanzimanyi:
Superseded by the consolidated PR branch `consolidate/tiles-bins-parity`. All changes from this PR are included in that branch as a single squashed commit. Please review and merge the consolidated branch instead.

PR #87, opened 2026-05-01
feat(bindings): SkipList-state aggregates — full *Agg surface (RFC #827)

estebanzimanyi:
## Summary

Builds the harness for MEOS SkipList-state temporal aggregates and ships seven of them.

| Aggregate | Inputs | Output |
|---|---|---|
| `tand`, `tor` | `tbool` | `tbool` |
| `tcount` | `tbool`, `tint`, `tfloat`, `ttext`, `tgeompoint` | `tint` |
| `tsum` | `tint`, `tfloat` | same as input |
| `tavg` | `tint`, `tfloat` | `tfloat` |
| `tcentroid` | `tgeompoint` | `tgeompoint` |

## How the harness works

SkipList aggregates differ from the previously-shipped fixed-state `extent()` family in two ways:

1. **State holds a `SkipList *` pointer.** The skiplist owns variable-size heap-allocated `Temporal *` values, so the aggregate needs a destructor — registered via `UnaryAggregateDestructor` with `AggregateDestructorType::LEGACY`.

2. **Combine merges via `temporal_tagg_combinefn`.** That function is exported from `libmeos.a` but **not declared in the public MEOS headers**, so we forward-declare it here alongside the per-aggregate `datum_*` merge functors.

The MEOS combine semantics share `Temporal *` pointers between the two input skiplists when both are non-empty: the returned skiplist owns the merged elements while the "loser" skiplist still references the same pointers. To avoid a double-free I walk the loser's elements and NULL out their `key`/`value` fields before calling `skiplist_free` — which then only releases the structural memory.

## Deferred

- **`tmin` / `tmax` aggregates.** Collide with the existing scalar `Tmin(tbox|stbox) → timestamptz` / `Tmax(tbox|stbox) → timestamptz` registrations under DuckDB's case-insensitive name resolution. DuckDB rejects mixing scalar+aggregate overloads on the same canonical name (raises `Not implemented Error: GetAlterInfo not implemented for this type` at LOAD time). Resolving needs either renaming the four box scalars (e.g. `boxTmin`/`boxTmax`) or suffixing the aggregates (e.g. `tMin_agg`) — out of scope for this PR.
- **`merge`, `appendInstant`, `appendSequence`** — same harness pattern, just different transfn names; can ship next.
- **`spanUnion`, `setUnion`** — need a different state shape (span/set skiplist instead of temporal).
- **Window aggregates** (`wmin`, `wmax`, `wsum`, `wcount`, `wavg`) — need an extra `interval` argument.
- **`tcentroid` 3D dispatch** — currently always uses `datum_sum_double3` (2D). Will need a runtime check on the `STBox` flags for 3D inputs.
- **`tcount` over scalar `timestamptz` / `tstzset` / `tstzspan` / `tstzspanset`** — uses `timestamptz_tcount_transfn` and friends; different signature.

## Test plan

- [x] `test/sql/parity/031_aggregates_skiplist.test` — 14 assertions, one per registered aggregate plus NULL / empty handling.
- [x] No regression in the existing `030_aggregates_extent.test`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,6 +71,7 @@ set(EXTENSION_SOURCES
     src/temporal/span.cpp
     src/temporal/span_functions.cpp
     src/temporal/span_aggregates.cpp
+    src/temporal/temporal_aggregates.cpp
     src/geo/geoset.cpp
     src/temporal/spanset.cpp
     src/temporal/spanset_functions.cpp
--- a/src/include/temporal/temporal_aggregates.hpp
+++ b/src/include/temporal/temporal_aggregates.hpp
@@ -1,0 +1,15 @@
+#pragma once
+
+#include "duckdb/main/extension/extension_loader.hpp"
+
+namespace duckdb {
+
+// SkipList-state temporal aggregates (tcount / tand / tor / tmin / tmax /
+// tsum / tavg / tcentroid). Registered separately from the fixed-state
+// extent() aggregates (in span_aggregates.cpp) because they need a
+// destructor and reach into MEOS skiplist internals.
+struct TemporalAggregates {
+    static void RegisterAggregateFunctions(ExtensionLoader &loader);
+};
+
+} // namespace duckdb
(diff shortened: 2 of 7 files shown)


estebanzimanyi:
Superseded by the consolidated PR branch `consolidate/aggregates-parity`. All changes from this PR are included in that branch as a single squashed commit. Please review and merge the consolidated branch instead.

PR #111, opened 2026-05-07
fix(threads): replace global mutex with per-thread MEOS initialization

estebanzimanyi:
> 👀 **Reviewers**: tier ranking, dependency chains and the standards checklist live in [`doc/contributing/reviewer-guide.md`](https://github.com/MobilityDB/MobilityDuck/blob/main/doc/contributing/reviewer-guide.md).

## Summary

- Replaces the global `std::mutex` from PR #51 with a `thread_local bool` guard in `EnsureMeosInitializedOnThread()`.
- Each DuckDB worker thread calls `meos_initialize()` + `meos_initialize_error_handler()` exactly once on first use.
- Cost: one bool check per scalar-function call — negligible; full DuckDB parallelism is preserved.

## Root cause

MEOS [PR #815](https://github.com/MobilityDB/MobilityDB/pull/815) made all mutable MEOS state thread-local (`MEOS_TLS`): the PROJ context, GSL RNGs, error code, and caches. The error handler is the only process-global piece.

`std::call_once` fires exactly once — on whichever thread first loads the extension. DuckDB's other worker threads never called `meos_initialize()`, so with MEOS's TLS design those threads hit NULL TLS pointers on any MEOS call and segfaulted on parallel aggregate queries.

The global mutex in PR #51 serialised every MEOS call through one thread (destroying DuckDB's parallelism) but still did not call `meos_initialize()` on worker threads — mutex ensures exclusion, not initialization.

## Test plan

- [x] Build succeeds with no warnings.
- [x] `tgeompointSeq(list(TGEOMPOINT(geom, ts) ORDER BY ts))` with `SET threads = 4`, 2000 rows / 50 vessels — returns correct results across worker threads.
- [x] All existing tests pass.

Closes #51

### DIFF
--- a/src/include/mobilityduck/meos_exec_serial.hpp
+++ b/src/include/mobilityduck/meos_exec_serial.hpp
@@ -1,38 +1,56 @@
 #pragma once
 
-#include <mutex>
-
 #include "duckdb/function/scalar_function.hpp"
 #include "duckdb/main/extension/extension_loader.hpp"
 
-namespace duckdb {
-
-/**
- * Mutex serialization for MobilityDuck scalar function bodies that ultimately call
- * MEOS / liblwgeom. MEOS uses GEOS's legacy global API (`initGEOS` / `finishGEOS`)
- * internally from worker threads; concurrent legacy GEOS init/teardown from DuckDB's
- * parallelism triggers intermittent GEOS errors and allocator corruption when spatial
- * and MobilityDuck coexist.
- *
- * Wrapping each MEOS-backed ScalarFunction execution prevents overlapping MEOS/GEOS
- * legacy pairs across concurrent pipelines while preserving parallelism elsewhere.
- */
-inline std::mutex &MeosSerializedExecMutex() {
-	static std::mutex mutex;
-	return mutex;
+extern "C" {
+void meos_initialize(void);
+void meos_initialize_error_handler(void (*)(int, int, const char *));
 }
 
-inline ScalarFunction WrapScalarFunctionWithMeosExecMutex(ScalarFunction sf) {
+namespace duckdb {
+
+// Defined in mobilityduck_extension.cpp — converts MEOS errors to DuckDB exceptions.
+extern "C" void MobilityduckMeosErrorHandler(int errlevel, int errcode, const char *errmsg);
+
+/**
+ * Ensure MEOS is initialised on the calling thread.
+ *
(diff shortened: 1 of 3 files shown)


estebanzimanyi:
## Coordination note — `main` has incompatible timezone commits

`main` has two commits that this PR will conflict with on merge:

- `39921f1 fix(tz): single-timezone model — extension forces both MEOS and DuckDB to Europe/Brussels` adds `meos_initialize_timezone("Europe/Brussels")` + `ExtensionHelper::AutoLoadExtension(db, "icu")` + `DBConfig::SetOptionByName("TimeZone", "Europe/Brussels")` to `LoadInternal` — exactly the block this PR's `c237f6c` removes.
- `08a5598 docs(tz): clarify two-timezone reality in comments` is comments framing the Brussels override.

Plus: the `update_test_expected.py` rewrite that landed alongside `39921f1` pinned ~25 test files to `+01` Brussels offset; once `9dd765a`'s timezone-neutral policy is the law of the land those need rewriting (to `=`/`stbox_eq`/`asText` comparisons), not flipped to `+00`.

Resolution path (maintainer's call — see `docs/CONSOLIDATION-PLAN.md` on #115): revert `39921f1` and `08a5598` from `main` either before or as part of this PR's merge, then rewrite the affected `+01` test expectations to timezone-neutral form. Some of that rewrite has already happened on PR #112 (`a10ef79`).

estebanzimanyi:
Closing — superseded by the MEOS-level thread-safety work in
MobilityDB PR #949 (\`feat/meos-thread-safe-complete\`).

This PR proposed a binding-side fix (per-thread \`meos_initialize\`
on each scalar-function call) for the same DuckDB-parallel-exec
SIGSEGV class.  The proper fix sits one layer down:

  - Per-thread GEOS context in MEOS + liblwgeom
  - Reentrant GEOS API across every spatial helper
  - JVM-safe (atomic) error handler
  - Threaded stress tests (\`meos/test/threaded_geos_test.c\`)

MobilityDuck's MEOS vcpkg port is now pinned to PR #949's tip
(\`742c1fb5\`) via the consolidated parity batch \`feat/parity-
additions-batch\` (PR #126), which restores the simpler
"one-time at extension load" init pattern that the new MEOS
supports correctly.

Branch \`fix/per-thread-meos-init\` left intact in case a
regression surfaces and the per-thread shim turns out to add
value on top of the upstream fix.

PR #116, opened 2026-05-10
consolidate: revert main's tz commits and align tests with TZ=UTC (unblocks PR #111)

estebanzimanyi:
> 👀 **Reviewers**: tier ranking, dependency chains and the standards checklist live in [`doc/contributing/reviewer-guide.md`](https://github.com/MobilityDB/MobilityDuck/blob/main/doc/contributing/reviewer-guide.md).

## Summary

Resolves the conflict between \`main\` and PR #111 (\`fix/per-thread-meos-init\`) so PR #111 can merge cleanly.

Three reverts on \`main\`:
- \`98bf131\` Revert \`39921f1 fix(tz): single-timezone model — extension forces both MEOS and DuckDB to Europe/Brussels\` — drops the \`AutoLoadExtension(icu)\` + \`SetOptionByName("TimeZone", ...)\` block from \`LoadInternal\`.
- \`3a23f08\` Revert \`08a5598 docs(tz): clarify two-timezone reality in comments\`.
- \`7adb3ee\` Revert \`47df9ca fix(tz): set MEOS timezone to Europe/Brussels for deterministic temporal I/O\` — drops \`meos_initialize_timezone("Europe/Brussels")\`.
- \`f733bd8\` Revert \`d3d9810 fix(tz): force MEOS UTC timezone at extension load; use Brussels in test runner\` — drops the earlier \`meos_initialize_timezone("UTC")\` so \`LoadInternal\` matches PR #111's parent state exactly.

Plus one mechanical follow-up:
- \`f5b2a2d test: shift remaining +01 expectations to +00\` — 23 test files where surviving \`+01\` expectations remained from earlier Brussels-pinned authoring.  Calendar dates unchanged; only the displayed offset shifts.

## Result

After this PR lands, \`LoadInternal\` is exactly:

\`\`\`cpp
static void LoadInternal(ExtensionLoader &loader) {
    ConfigureMeosSridCsvOnce();
    static std::once_flag meos_init_flag;
    std::call_once(meos_init_flag, []() {
        meos_initialize();
        meos_initialize_error_handler(&MobilityduckMeosErrorHandler);
    });
    // ...
}
\`\`\`

PR #111 then replaces the \`std::call_once\` block with \`EnsureMeosInitializedOnThread()\` — a clean diff with no policy fight.

## Long-term direction

[...]

PR #118, opened 2026-05-10
fix: sync MEOS API drift (meosType→MeosType + spatial-rel arg drop)

estebanzimanyi:
## Summary

Two cherry-picked commits unblocking local compilation against the current vcpkg MEOS. `make release` on `main` currently fails with `'meosType' does not name a type; did you mean 'MeosType'?` and `too many arguments to function 'Temporal* tcontains_geo_tgeo(...)'` (× 13 call sites); without these patches no parity work builds locally, and only branches that already carry the same fix get green CI.

- `fe4a0da refactor: meosType -> MeosType (sync to upstream MEOS PR #785)` — sweeps every reference of the lowercase `meosType` typedef in MobilityDuck source to the PascalCase `MeosType` that the current MEOS catalog header exposes.
- `cd1d5d2 fix(geo): adapt 12 t{contains,disjoint,intersects,touches,dwithin}_* call sites to 3-arg MEOS API` — the `tcontains_geo_tgeo`, `tdisjoint_*`, `tintersects_*`, `ttouches_*`, `tdwithin_*` exports no longer carry the `restr` / `at_value` trailing arguments (the result is the full tbool; restriction is composed at the call site via `tbool_at_value` when needed). Updates the 12 call sites in `tgeompoint_functions.cpp`.

Both commits originate from the parallel `perf/duck-stack-attempt2` work (`7a8cc22`, `c37b9e2`); cherry-picked onto `main` so they land independently and other in-flight branches stop compiling against a stale API.

## Test plan

- [ ] CI green on `linux_amd64` and `linux_arm64`.
- [ ] No behavioural change — the only places `restr=true` was used are wrapped in a `tbool_at_value(ret, at_value)` post-restriction that produces the same result the dropped MEOS args used to compute internally.

### DIFF
--- a/src/geo/tgeompoint_functions.cpp
+++ b/src/geo/tgeompoint_functions.cpp
@@ -454,7 +454,7 @@ void TgeompointFunctions::Tgeompoint_sequence_constructor(DataChunk &args, Expre
     
     auto arg_count = args.ColumnCount();
     auto row_count = args.size();
-    meosType temptype = TemporalHelpers::GetTemptypeFromAlias(result.GetType().GetAlias().c_str());
+    MeosType temptype = TemporalHelpers::GetTemptypeFromAlias(result.GetType().GetAlias().c_str());
     interpType interp = temptype_continuous(temptype) ? LINEAR : STEP;
     bool lower_inc = true;
     bool upper_inc = true;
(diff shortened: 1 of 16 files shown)


estebanzimanyi:
Superseded by #121 `main hygiene: revert single-timezone + sync MEOS API drift (consolidates #116 + #118)`. No content lost; cherry-picks applied cleanly. Close this PR when convenient.

estebanzimanyi:
Folded into #121 (same) per the ecosystem PR consolidation policy.  The consolidation PR explicitly declares this PR as one of its inputs in the `Closes` line; verified content equivalence (byte-identical / additive refinements only).  Closing as superseded; no work is lost.

PR #115, opened 2026-05-10
docs: PR coordination policy + consolidation plan

estebanzimanyi:
> 👀 **Reviewers**: tier ranking, dependency chains and the standards checklist live in [`doc/contributing/reviewer-guide.md`](https://github.com/MobilityDB/MobilityDuck/blob/main/doc/contributing/reviewer-guide.md).

## Summary

Two new docs covering the ecosystem-coordination policy that the open consolidation work depends on:

- `docs/PR-COORDINATION.md` — ecosystem policy: `gh pr list` is the first step before any code change. Don't duplicate work, don't conflict with in-flight consolidations, don't reintroduce policies that an open PR is reverting.
- `docs/CONSOLIDATION-PLAN.md` — working artefact mapping the file-level overlap between commits already on `main` and the open `consolidate/*` parity PRs (#97/#98/#99/#100/#102/#103/#104) plus PR #111 (per-thread MEOS init). Three resolution options per overlap (rebase / revert+fold / keep both); maintainer picks per PR.

The policy doc spells out the rule; the consolidation plan documents the live overlaps so the maintainer has a clear picture before reconciling them. The two failure modes the policy prevents:

1. Six commits on `main` (`e41c8d9 → c8cad6d`) duplicating surface that the open `consolidate/*` PRs already cover.
2. `39921f1 fix(tz): single-timezone model` baking in a `meos_initialize_timezone(...)` + `SetOptionByName(...)` policy that PR #111 (`fix/per-thread-meos-init`) is simultaneously reverting in favour of per-thread MEOS init plus timezone-neutral test assertions (commit `9dd765a`).

## Test plan

- [ ] Docs render correctly on the rendered branch.
- [ ] No code changes — should be green on every CI step that doesn't fail trivially.

PR #119, opened 2026-05-10
scripts: add lint for timezone-pinned test expected values

estebanzimanyi:
## Summary

`scripts/lint-tz-pinned-tests.py` flags every line in an expected-output block (after `----`) that carries a hardcoded UTC offset.  Output is `file:line: snippet` for each hit, with a closing summary count and a pointer to the timezone-neutral patterns to replace them with.

Designed as a pre-commit gate or a CI lint step.  No code changes anywhere else; purely a maintainer-facing tool.

## Why

PR #111 / commit \`9dd765a test(stbox): make timestamp assertions timezone-neutral\` established the policy that test assertions should not pin a timezone offset, because MEOS renders timestamps using the process's TZ and any divergence between developer machine and CI flips the expected offset.  Today there are 734 such pinned values across 43 test files (current snapshot of \`main\`).  The lint makes that surface visible and gives the maintainer a tool to prevent regressions while the existing pins are progressively rewritten.

## Test plan

- [x] Running on \`main\` reports 734 hits (matches `grep -rE '^[+-][0-9]{2}\b' test/`).
- [x] Excludes scientific-notation false positives (`1.5e+00`).
- [ ] CI green.

## Related

- #111 \`fix/per-thread-meos-init\` — establishes the timezone-neutral policy.
- #116 \`consolidate/tz-revert-and-neutralize\` — interim shift +01→+00 to unblock #111; the lint catches the long-term work that follow-up PRs will do.

### DIFF
--- a/scripts/lint-tz-pinned-tests.py
+++ b/scripts/lint-tz-pinned-tests.py
@@ -1,0 +1,109 @@
+#!/usr/bin/env python3
+"""Flag timezone-pinned expected values in DuckDB sqllogic tests.
+
+A test that hardcodes a timezone offset (`+00`, `+01`, `+02`, …) in an
+expected output line is fragile: MEOS renders timestamps using the
+process's TZ, and any divergence between the developer's machine and CI
+flips the expected offset.  The project policy is to write
+timezone-neutral assertions instead — value equality (`= tstzspan
+'...'`), accessor functions (`stbox_eq`, `numSpans`, `numValues`,
+`startTimestamp() = …`), or `asText(...)` round-trips on both sides.
+
+This script walks `test/sql/**/*.test`, finds every line in an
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
Superseded by #120 `docs+scripts: PR coordination policy + tz-pinned-test lint (consolidates #115 + #119)`.  No content lost; cherry-picks applied cleanly with no conflicts.  Close this PR when convenient.

estebanzimanyi:
Folded into #120 (same) per the ecosystem PR consolidation policy.  The consolidation PR explicitly declares this PR as one of its inputs in the `Closes` line; verified content equivalence (byte-identical / additive refinements only).  Closing as superseded; no work is lost.

PR #126, opened 2026-05-10
feat(parity): batch — bearing + eCovers/tCovers + stbox dim + seqSetGaps (consolidates #122+#123+#124+#125)

estebanzimanyi:
> 👀 **Reviewers**: tier ranking, dependency chains and the standards checklist live in [`doc/contributing/reviewer-guide.md`](https://github.com/MobilityDB/MobilityDuck/blob/main/doc/contributing/reviewer-guide.md) (lands with [PR #117](https://github.com/MobilityDB/MobilityDuck/pull/117)).

## Summary

Rolling-topic parity additions PR per the ecosystem PR consolidation policy. Four genuinely independent parity additions to MobilityDuck, each one feature commit, no file overlap between commits:

- **`feat(parity): bearing — initial bearing for tgeompoint and tgeogpoint`** (was [PR #122](https://github.com/MobilityDB/MobilityDuck/pull/122))
  Adds `bearing(tgeompoint, tgeompoint)` and `bearing(tgeogpoint, tgeogpoint)` mapping to MEOS's `bearing_*` family.

- **`feat(parity): eCovers (BOOLEAN) and tCovers (tbool) for tgeo`** (was [PR #123](https://github.com/MobilityDB/MobilityDuck/pull/123))
  Closes the asymmetric gap on the spatial-rel surface — `eCovers` returns BOOLEAN (ever-covers), `tCovers` returns tbool (per-instant covers).

- **`feat(parity): stbox dimensional constructors (stboxX/Z/T/XT/ZT, geodstbox*)`** (was [PR #124](https://github.com/MobilityDB/MobilityDuck/pull/124))
  Exposes the 6 dimension-explicit STBOX constructors plus the geodetic variants — matches MobilityDB's surface.

- **`feat(parity): SeqSetGaps for all 8 temporal types (closes long-standing #187)`** (was [PR #125](https://github.com/MobilityDB/MobilityDuck/pull/125))
  Adds `seqSetGaps(temporal, interval)` for tbool/tint/tfloat/ttext/tgeompoint/tgeogpoint/tgeometry/tgeography.  Closes the long-standing #187 user request.

The four were originally split because they're independent parity additions (different MEOS surfaces, different test files).  Folding them is purely a queue-optimisation: same domain (parity additions to MobilityDuck), same author, zero file overlap (each adds its own `0NNb_*.test` file under `test/sql/parity/`).

[...]

PR #128, opened 2026-05-10
fix(vcpkg): bump MEOS port pin to current MobilityDB master

estebanzimanyi:
## Summary

Bumps `vcpkg_ports/meos/portfile.cmake` `REF` from `f11b7443e` (Mar 30) to `ee27da1a6` (current `master`).  The newer MEOS exposes the catalog-type enum as `MeosType` (PascalCase) and ships the spatial-rel exports (`tcontains_geo_tgeo`, `tdisjoint_*`, `tintersects_*`, `ttouches_*`, `tdwithin_*`, `tpoint_minus_geom`) in their orthogonal 2-/3-arg form.

The MobilityDuck source already uses these names and signatures; the older pin made CI fail with `'MeosType' does not name a type` and `too many arguments to function 'Temporal* tcontains_geo_tgeo(...)'`.

## Test plan

- [ ] Linux amd64 + arm64 CI compiles MobilityDuck against the new MEOS pin
- [ ] `feat/parity-additions-batch` (PR #126) CI turns green once this lands

### DIFF
--- a/vcpkg_ports/meos/portfile.cmake
+++ b/vcpkg_ports/meos/portfile.cmake
@@ -1,8 +1,8 @@
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO MobilityDB/MobilityDB
-    REF f11b7443ee985dc1ffb778c325e62f0edaf255ec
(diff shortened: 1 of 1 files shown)


estebanzimanyi:
Superseded by PR #126 — the port bump alone fails to compile because `main` lacks the drift fixes for spatial-rel exports.  Folded the bump into PR #126 (commit `215cb79`) so the bump and the drift fix land together.  Branch `fix/vcpkg-meos-pin-bump` will be deleted.

PR #132, opened 2026-05-13
spaceSplit and spaceTimeSplit accept LATERAL column inputs

estebanzimanyi:
Re-register both splitters as in_out_function table functions so per-row parameters (tgeompoint, sizing constants, origin, duration, torigin, bitmatrix) are read at exec time from the input DataChunk rather than bound once from literals; pattern mirrors duckdb/extension/icu/icu-table-range.cpp. When the input chunk drains, the function leaves output.size() at zero and returns NEED_MORE_INPUT, matching PhysicalTableScan's FINISHED-by-chunk-size contract and PhysicalTableInOutFunction's NEED_MORE_INPUT advance. test/sql/parity/058c_tpoint_split_lateral.test exercises the LATERAL column-input path alongside 058b's literal-input regression. Stacked on #126, which carries the MEOS pin bump and API adaptations the build depends on.

### DIFF
--- a/src/geo/tgeompoint.cpp
+++ b/src/geo/tgeompoint.cpp
@@ -2074,28 +2074,32 @@ void TgeoFromTextExec(DataChunk &args, ExpressionState &state, Vector &result) {
 
 namespace {
 
+/* spaceSplit / spaceTimeSplit support both literal and LATERAL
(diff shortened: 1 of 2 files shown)


estebanzimanyi:

**What this PR does:** spaceSplit and spaceTimeSplit accept LATERAL column inputs.

PR #133, opened 2026-05-13
Drop the unnecessary *Agg suffix from non-colliding aggregates

estebanzimanyi:
DuckDB's catalog dispatches on argument types, so the same-stem collision that previously motivated the *Agg suffix (RFC #827, e.g. aggregate Tmin on tnumber vs scalar Tmin on tbox) is resolved by signature dispatch: the aggregate takes a temporal value, the scalar takes a box, they coexist. The suffix read as ugly and broke the principle that the cross-platform SQL form should match what a Snowflake/Databricks/lakehouse programmer would naturally write. All 21 aggregates and their test references are renamed (TandAgg to Tand, TminAgg to Tmin, MergeAgg to Merge, WavgAgg to Wavg, etc.); doc comments in src/temporal/temporal_aggregates.cpp and the four affected parity test headers replace the *Agg-convention prose with the simpler argument-type-dispatch explanation. Pure rename plus doc cleanup, no semantic change. CI is currently blocked by an unrelated pre-existing vcpkg/MEOS pin issue on origin/main: portfile REF f11b7443e (pre-Apr-28 meosType to MeosType rename) does not match SHA512 ae8589acc8 (which corresponds to 742c1fb5, post-rename), so fresh downloads compile-fail at src/include/tydef.hpp:18 with "MeosType does not name a type"; a separate vcpkg-pin-fix PR is the right vehicle since fully bumping MEOS cascades into per-call signature updates beyond this rename scope.

### DIFF
--- a/src/geo/spatial_aggregates.cpp
+++ b/src/geo/spatial_aggregates.cpp
@@ -281,7 +281,7 @@ void SpatialAggregates::AddExtentOverloads(AggregateFunctionSet &extent_set) {
 }
 
 void SpatialAggregates::RegisterTcentroid(ExtensionLoader &loader) {
-    AggregateFunctionSet tcentroid_set("TcentroidAgg");
+    AggregateFunctionSet tcentroid_set("Tcentroid");
     tcentroid_set.AddFunction(MakeTcentroidAggregate(TgeompointType::TGEOMPOINT()));
     loader.RegisterFunction(std::move(tcentroid_set));
 }
--- a/src/temporal/set.cpp
+++ b/src/temporal/set.cpp
@@ -1067,7 +1067,7 @@ struct SetUnionSetFunction {
 } // anonymous namespace
 
 void SetTypes::RegisterSetUnionAgg(ExtensionLoader &loader) {
-    AggregateFunctionSet set_union_set("SetUnionAgg");
+    AggregateFunctionSet set_union_set("SetUnion");
 
     // Scalar overloads: convert each value to a single-element Set.
     set_union_set.AddFunction(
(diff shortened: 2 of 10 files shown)


estebanzimanyi:

**What this PR does:** Drop the *Agg suffix from all aggregate names.

estebanzimanyi:
Aggregates whose bare name collides with a scalar must keep a disambiguating suffix: DuckDB maps a function name to a single catalog entry kind, so registering an aggregate over the existing scalar time-min/max accessors (Tmin/Tmax on tbox/stbox) fails extension load with 'GetAlterInfo not implemented'. The aggregate surface registers TminAgg/TmaxAgg (and the rest of the *Agg set) for this reason.

estebanzimanyi:
Reopening: the *Agg suffix is a binding-surface workaround that breaks the ecosystem auto-codegen (all bindings are mechanical remaps of the canonical MEOS name). The tmin/tmax scalar-accessor vs temporal-aggregate name collision is resolved at the MEOS/canonical level so every binding inherits the same distinct names, not by suffixing in DuckDB alone.

estebanzimanyi:
Superseded by MobilityDB #828 (green): the scalar/aggregate name collisions are resolved canonically at the MEOS surface by suffixing the aggregate form (mergeAgg/appendInstantAgg/appendSequenceAgg/minDistanceAgg/tminAgg/tmaxAgg), which every binding inherits via codegen. MobilityDuck already registers TminAgg/TmaxAgg, so it conforms; dropping the suffix would re-introduce the load collision.

PR #135, opened 2026-05-15
Expose minDistance scalar UDFs

estebanzimanyi:
Registers minDistance(tgeompoint, geometry), minDistance(geometry, tgeompoint), and minDistance(tgeompoint, tgeompoint) following the nearestApproachDistance pattern. The geo-typed forms reuse the NAD kernel since NAD reduces to spatial-min when one argument has no time dimension; the tgeompoint/tgeompoint form calls the threshold-aware mindistance_tgeo_tgeo kernel. The canonical cross-join shape is MIN(minDistance(t1, t2)) with GROUP BY, the cloud-SQL-natural Solution-2 form, identical in result to the MobilityDB 2-ary aggregate of the same signature. Stacked on the MEOS pin bump branch; advances the pin one further commit to d4cc7d289 so the mindistance kernels are present, a commit that only adds mindistance_tgeo_tgeo and tgeoarr_tgeoarr_mindist and changes no existing signature.


estebanzimanyi:

**What this PR does:** Expose minDistance scalar UDFs.

PR #134, opened 2026-05-14
Bump pinned MEOS and adapt to orthogonalized spatial-rels API

estebanzimanyi:
Bumps the vcpkg portfile to MEOS dd4ccd3c2 (post-rename + post-spatial-rels orthogonalization) and adapts the temporal-spatial relationship surfaces (tContains, tDisjoint, tIntersects, tTouches, tDwithin) to the new MEOS API which no longer takes the restr/at_value parameters; the previous behavior is reproduced by composing with atValues(temporal, bool). Also drops the now-defunct 3-arg minusGeometry(..., FLOATSPAN) overload since tpoint_minus_geom no longer accepts a z-span.

PR #136, opened 2026-05-15
Pre-stage icu extension for amd64 docker tests

estebanzimanyi:
MobilityDuck's LoadInternal calls `ExtensionHelper::AutoLoadExtension(db, "icu")` so the Europe/Brussels timezone option is honoured. Inside the linux_amd64 test docker container there is no network egress for the extension hub and the local extension directory is empty, so the autoload fails with `Extension "…/v1.4.4/linux_amd64/icu.duckdb_extension" not found`. The icu artifact is built locally as part of this extension's build (declared in extension_config.cmake); this commit copies it into the expected on-disk path before running the unittester so autoload resolves without a network round-trip. linux_arm64 has no test step in the upstream workflow so the asymmetry between archs was masked.


estebanzimanyi:

**What this PR does:** Pre-stage icu extension for amd64 docker tests.

**Files to read:** `` (plus minimal surrounding context).

estebanzimanyi:

PR #137, opened 2026-05-16
Initialize MEOS per worker thread

estebanzimanyi:
The MEOS 1.4 uplift keeps the session timezone, errno, PROJ context and RNGs in thread-local storage and requires every thread that calls into MEOS to run meos_initialize() before its first call. The extension only initialized MEOS once on the load thread via std::call_once, so DuckDB TaskScheduler worker threads executed scalar, cast and aggregate bodies with a NULL session_timezone and segfaulted in pg_next_dst_boundary on the first timestamp parse (reproduced locally as a SIGSEGV in 042_tgeogpoint_parity, previously masked because amd64 CI never reached the tests and arm64 runs no test step). A thread-local guard now runs the per-thread init and re-installs the process-global MobilityDuck error handler (meos_initialize() resets it to the exit-on-error default), invoked from the scalar exec wrapper and through a cast registration trampoline that covers every cast entry point. Locally the full suite goes from unrunnable to 58/59 files and 1311/1312 assertions with no segfaults; the one remaining assertion is an unrelated pre-existing ln() output drift from the MEOS bump.

### DIFF
--- a/src/geo/geoset.cpp
+++ b/src/geo/geoset.cpp
@@ -36,12 +36,12 @@ void SpatialSetType::RegisterTypes(ExtensionLoader &loader){
 }
 
 void SpatialSetType::RegisterCastFunctions(ExtensionLoader &loader) {        
-    loader.RegisterCastFunction(
+    RegisterMeosCastFunction(loader, 
         LogicalType::VARCHAR, 
         SpatialSetType::geomset(),                                    
         SpatialSetFunctions::Text_to_geoset   
     );     
-    loader.RegisterCastFunction(
+    RegisterMeosCastFunction(loader, 
         LogicalType::VARCHAR, 
         SpatialSetType::geogset(),                                    
         SpatialSetFunctions::Text_to_geoset   
--- a/src/geo/stbox.cpp
+++ b/src/geo/stbox.cpp
@@ -27,43 +27,43 @@ void StboxType::RegisterType(ExtensionLoader &loader) {
 }
 
 void StboxType::RegisterCastFunctions(ExtensionLoader &loader) {
-    loader.RegisterCastFunction(
+    RegisterMeosCastFunction(loader, 
         LogicalType::VARCHAR,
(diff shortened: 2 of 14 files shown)


estebanzimanyi:
## Reviewer's quickstart — ~8 minutes

**What this PR does in one sentence:** moves `meos_initialize()` from process-once to **per-DuckDB-worker-thread**, fixing the data-race where MEOS's thread-local state (timezone, error handler, lwgeom session) was uninitialised in workers spawned after the initial `LoadInternal`.

**Files (14):** mostly `src/geo/*.cpp` and `src/temporal/*.cpp` — each scalar/cast/cast-trampoline now starts with `MeosThread::ensureReady()` (a TLS-guarded one-time init).

**Read first:**
- **`src/include/mobilityduck/meos_thread.hpp`** (new) — the TLS guard + `ensureReady()` definition.
- **`src/include/mobilityduck/meos_exec_serial.hpp`** (new) — the serial-execution wrapper for paths MEOS itself isn't thread-safe on (pg-derived TZ, legacy GEOS).
- **`src/mobilityduck_extension.cpp`** — registers the worker thread's MEOS state.
- 11 callsite files just add `MeosThread::ensureReady();` at the top of each UDF lambda body.

**Verification:**
```bash
# Repro pre-PR: SIGSEGV on multi-threaded queries
duckdb -unsigned -c "LOAD mobilityduck; SET threads=4; SELECT count(*) FROM (SELECT atTime(traj::VARCHAR, '2026-01-01'::TIMESTAMP) FROM ...);"
# Pre-PR: occasionally crashes. Post-PR: clean.
```

**Cross-link:** Linux arm64 needs **#161** for the `MeosType` build error.

**Why it's safe to merge:** pure additive guards; valid-input paths are unaffected. The change correctly diagnoses MEOS's process-singleton initialization model as a multi-worker bug.

estebanzimanyi:

PR #139, opened 2026-05-16
Fix TRTREE index crash on temporal and span columns

estebanzimanyi:
CREATE INDEX USING TRTREE on a tgeompoint or any non-stbox column crashed with DuckDB's generic "assertion failure within DuckDB" message because bind rejected anything but STBOX/TSTZSPAN through an InternalException and the create-physical Finalize path memcpy-reinterpreted the temporal's MEOS-WKB blob as an STBox struct. Bind now accepts stbox, tbox, the five span types and the eight temporal types and selects the matching MEOS R-tree flavour; Construct dispatches per column type, deriving an SRID-stripped STBox via tspatial_to_stbox for spatial temporals and using rtree_insert_temporal otherwise, while span and box blobs are inserted as-is. The create-physical parallel path delegates to the same Construct so index keys and the scan-time query box are derived identically, and an unsupported column type raises a clean BinderException instead of a fake internal crash. Adds test/sql/parity/050_index_types.test covering every supported column type end-to-end plus the unsupported-type regression. Stacked on #134 (fix/bump-meos-pin) whose post-bump MeosType and MeosArray-based rtree_search API this builds against; rebase onto main once #134 merges.

### DIFF
--- a/src/include/index/rtree_module.hpp
+++ b/src/include/index/rtree_module.hpp
@@ -86,6 +86,7 @@ private:
 
     MeosType bbox_meostype;
     size_t bbox_size_;
+    LogicalType column_type_;
 
     size_t current_size_ = 0;
     size_t current_capacity_ = 0;
--- a/src/index/rtree_index_create_physical.cpp
+++ b/src/index/rtree_index_create_physical.cpp
@@ -151,11 +151,9 @@ public:
 
 			const auto count = scan_chunk.size();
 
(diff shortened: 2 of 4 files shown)


estebanzimanyi:

**What this PR does:** Fix TRTREE index crash on temporal and span columns.

**Files to read:** `src/index/` (plus minimal surrounding context).

PR #138, opened 2026-05-16
Refresh tnumber ln/exp/log expected output for MEOS turning points

estebanzimanyi:
The MEOS uplift adds tfloat_ln_turnpt / tfloat_exp_turnpt (MobilityDB #1003), which insert one chord-error-minimising turning point on a linear input segment for the transcendental unary lifts, so ln/log10/exp over a two-instant tfloat now return three instants. This refreshes the 026b expectations to the new values (captured from the built extension); deltaValue, trend and the arithmetic-alias cases are linear and unchanged. With this on top of the per-thread MEOS init fix the amd64 suite is fully green (59/59 files locally and the previously failing assertion now passes).

PR #140, opened 2026-05-16
Wrap bigint_to_set in an int64_t forwarder for SetUnionAgg

estebanzimanyi:
On macOS LP64 int64 (long) and int64_t (long long) are the same width but distinct types, so clang rejects bigint_to_set as the non-type template argument of SetUnionScalarFunction<int64_t, ...>, failing the osx_amd64 and osx_arm64 builds in the icu-fix stack leaf #138. This adds a bigint_to_set_duckdb forwarder taking int64_t and casting to int64 before calling MEOS, mirroring the existing date_to_set_duckdb idiom; the cast is a no-op on Linux. Stacked on #138 (fix/tnumber-math-turnpt-expected); its commits show in this diff until it merges.

### DIFF
--- a/src/temporal/set.cpp
+++ b/src/temporal/set.cpp
@@ -945,6 +945,13 @@ static inline Set *date_to_set_duckdb(DateADT d) {
     return date_to_set(ToMeosDate(duckdb::date_t(d)));
 }
 
+// macOS LP64: int64 (long) and int64_t (long long) are distinct types, so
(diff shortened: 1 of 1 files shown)


estebanzimanyi:

**What this PR does:** Wrap bigint_to_set in an int64_t forwarder for SetUnionAgg.

**Files to read:** `src/include/temporal/set.hpp` (plus minimal surrounding context).

estebanzimanyi:

PR #141, opened 2026-05-16
Stage icu for the macOS osx_arm64 test path too

estebanzimanyi:
The stage_icu helper mapped only the Linux uname values, so on the macOS arm64 test runner uname -m returned arm64 and the locally built icu extension was copied to .duckdb/extensions/v1.4.4/arm64 instead of .../osx_arm64 where DuckDB autoload looks, and the hub fallback is not reliably resolvable there, so the osx_arm64 Test step failed to load the extension. This maps the OS and architecture to the DuckDB platform string so icu is staged where autoload expects on every tested platform, with the Linux mapping unchanged. It is logically an iteration of the #136 icu pre-stage (build/CI wiring); stacked on #140 so the leaf goes fully green, and the committer may prefer to fold it into #136 at merge. Stacked on fix/macos-int64-bigint-to-set-forwarder (#140); its commit shows in this diff until it merges.


estebanzimanyi:

**What this PR does:** Stage icu for the macOS osx_arm64 test path too.

**Files to read:** `` (plus minimal surrounding context).

PR #142, opened 2026-05-16
Bump pinned MEOS to include the wasm pg_config fix

estebanzimanyi:
MobilityDuck builds MEOS from a pinned MobilityDB commit; the current pin predates the fix that emits SIZEOF_LONG_LONG, so the DuckDB-Wasm jobs fail in the vendored postgres pg_bitutils.h on wasm32. This advances the pin to a MobilityDB commit that carries that fix (MobilityDB PR #1036). It is provisionally pointed at estebanzimanyi/MobilityDB@40145502c because #1036 is not yet on MobilityDB/MobilityDB master, kept as a Draft for that reason; the landable form points REPO back to MobilityDB/MobilityDB at the post-merge master commit with SHA512 recomputed from that archive. Stacked on fix/stage-icu-macos-platform (#141); its commits show in this diff until it merges.


estebanzimanyi:

**What this PR does:** Bump pinned MEOS to include the wasm pg_config fix.

**Files to read:** `vcpkg_ports/meos/portfile.cmake` (plus minimal surrounding context).

estebanzimanyi:

PR #143, opened 2026-05-16
Use MEST multi-entry indexing in the TRTREE index

estebanzimanyi:
A temporal column was indexed as a single minimum bounding box per row, so a wiggly or high-extent trajectory matched far more query boxes than it actually overlapped. Construct now indexes each temporal value as up to max_boxes tight per-segment bounding boxes via the MEOS rtree_insert_temporal_split (degenerating to the previous single-box behaviour for instants or max_boxes <= 1); temporal-spatial values are SRID-normalised to 0 before splitting so every produced stbox matches the SRID-stripped query box, and Search deduplicates row ids with a seen-set because multi-entry leaves return the same id once per overlapping segment box. The split bound is exposed as WITH (max_boxes = N), default 8, validated on the MEST selectivity benchmark (false-positive rate 47.2 percent single-box to 13.8 percent at 8). The portfile.cmake change here pins MEOS at the MEST commit only as a TEMPORARY BUILD-BASE so this PR is green in isolation; it deliberately conflicts with the wasm pin work in #142. At finalization this PR drops the portfile edit entirely and inherits #142s pin, which requires MobilityDB #1032 (MEST) to be merged to master alongside #1036 (wasm) so a single master commit carries both; the MobilityDuck merge order is #134, #136, #137, #138, #140, #141, #142, then #139, then this PR. Stacked on the TRTREE crash fix (fix/trtree-index-assertion).

### DIFF
--- a/src/include/index/rtree_module.hpp
+++ b/src/include/index/rtree_module.hpp
@@ -87,6 +87,11 @@ private:
     MeosType bbox_meostype;
     size_t bbox_size_;
     LogicalType column_type_;
+    // Multi-entry (MEST) split bound for temporal columns: each temporal
+    // value is indexed as up to this many tight per-segment bounding
+    // boxes. <= 1 degenerates to the single minimum bounding box (the
+    // pre-MEST behaviour). Tunable via WITH (max_boxes = N) on the index.
+    int max_boxes_ = 8;
 
     size_t current_size_ = 0;
     size_t current_capacity_ = 0;
--- a/src/index/rtree_index_scan.cpp
+++ b/src/index/rtree_index_scan.cpp
@@ -28,12 +28,17 @@ BindInfo TRTreeIndexScanBindInfo(const optional_ptr<FunctionData> bind_data_p) {
 struct RTreeIndexScanGlobalState : public GlobalTableFunctionState {
 	DataChunk all_columns;
 	vector<idx_t> projection_ids;
+	vector<LogicalType> scanned_types;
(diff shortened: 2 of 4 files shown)


estebanzimanyi:

**What this PR does in one sentence:** switches the TRTREE index from single-entry per row to **multi-entry indexing (MEST)** — every spatial bbox of a temporal trajectory's component segments becomes an index entry, dramatically improving recall for spatial predicates.

**Files (5):**
- **`src/index/rtree_index_create_physical.cpp`** — the new MEST build path.
- **`src/index/rtree_index_scan.cpp`** — adjusted scan to dedup multi-entry hits.
- **`src/include/index/rtree_module.hpp`** — type declarations for the new index payload.
- **`test/sql/index/`** — assertions on recall (no false negatives) and result-set deduplication.

**Why it matters:** previously a single trajectory got ONE bbox in the index — that bbox is loose (covers the whole trajectory's extent), so queries with small predicate polygons returned huge false-positive sets that had to be re-filtered. With MEST, each segment's bbox is indexed separately; a small predicate polygon hits only the relevant segments.

**Foundational for:** #144 (constant-geometry pushdown — depends on the multi-entry structure existing).

**Cross-link:** Linux arm64 build needs #161.

**Why it's safe to merge:** the index becomes strictly more precise — never lower recall (every old entry still maps to the trajectory it came from), never wrong results. Build cost goes up (more entries), query cost goes down (less false-positive filtering).

PR #144, opened 2026-05-16
Route constant-geometry spatial-relationship predicates to the TRTREE index

estebanzimanyi:
eIntersects, eContains, eDisjoint and eTouches against a constant geometry are recognised by the TRTREE scan optimizer, which synthesizes the bounding-box && prefilter from the constant and probes the index (single-box or MEST). The bbox is a lossy superset, so the index scan reports supports_pushdown_type = false and DuckDB keeps the original predicate as an exact recheck filter above the scan, mirroring the lossy-index-always-rechecks contract of PostGIS GiST and MobilityDB's tspatial_supportfn. A parity block proves that a row whose bounding box overlaps the query polygon but does not actually intersect it is dropped, identically with and without the index and under MEST multi-entry. Stacked on #143; its commits show in the diff until it merges.

### DIFF
--- a/src/index/rtree_index_scan.cpp
+++ b/src/index/rtree_index_scan.cpp
@@ -125,14 +125,30 @@ static void RTreeIndexScanExecute(ClientContext &context, TableFunctionInput &da
 //-------------------------------------------------------------------------
 // Get Function
 //-------------------------------------------------------------------------
+
+// The R-tree probe narrows rows by bounding-box overlap only. For an exact
+// predicate (&&, @>) that is the answer; for a lossy predicate (the
+// spatial-rel functions, whose bbox is only a superset) the original
+// predicate must still be evaluated or the scan would emit false positives.
+// Reporting that this scan can apply no pushed expression filter makes
+// DuckDB keep every pushed predicate as an exact recheck PhysicalFilter
+// directly above the scan (execution/physical_plan/plan_get.cpp rebuilds it
+// via ExpressionFilter::ToExpression). This is the lossy-index-always-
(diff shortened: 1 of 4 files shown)


estebanzimanyi:
## Reviewer's quickstart — ~6 minutes

**What this PR does in one sentence:** when a spatial-rel predicate has a constant geometry on one side (e.g. `eIntersects(tgeompoint, 'POLYGON(...)'::GEOMETRY)`), routes the lookup through the TRTREE multi-entry index instead of scanning every row — proper index pushdown for the most common geofence query shape.

**Files (4):**
- **`src/index/rtree_optimize_scan.cpp`** — the scan-optimisation pass that recognises the constant-geometry pattern and rewrites the plan node.
- **`src/index/rtree_index_scan.cpp`** — the new constant-geometry-aware scan entry point.
- **`src/include/index/rtree_module.hpp`** — public declarations for the optimiser hook.
- **`test/sql/index/`** — assertions that the EXPLAIN plan uses the index for the new pattern.

**Verification:**
```sql
CREATE INDEX rt ON trips USING TRTREE (traj);
EXPLAIN SELECT count(*) FROM trips WHERE eIntersects(traj, 'POLYGON(...)'::GEOMETRY);
-- Before this PR: SEQ_SCAN.  After: TRTREE_INDEX_SCAN.
```

**Cross-link:** stacked on #143 (MEST multi-entry indexing), which is in turn the foundation for this optimisation. Both will rebuild cleanly once #161 lands (Linux arm64 build).

**Why it's safe to merge:** the pre-existing scan path remains available for non-constant-geometry cases; this only adds a new index-aware branch.

PR #146, opened 2026-05-16
Add temporalFooter for TemporalParquet KV_METADATA

estebanzimanyi:
temporalFooter builds the TemporalParquet footer-metadata JSON manifest for COPY ... TO '*.parquet', faithful to the ratified opaque-MEOS-WKB spec. The test covers the deterministic footer assertions and the KV_METADATA round-trip. Stacked on #134 (fix/bump-meos-pin) because current main pins a pre-MeosType MEOS that no longer compiles; base retargets to main once #134 lands. End-to-end MEOS-WKB round-trip and native-scalar sidecar pushdown coverage are deferred to a follow-up gated on the MobilityDuck binding-stability fix, since exercising many sequential MEOS calls in one process trips a pre-existing non-deterministic native crash tracked in a separate lane.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,6 +94,7 @@ set(EXTENSION_SOURCES
     src/index/rtree_index_create_physical.cpp
     src/index/rtree_index_scan.cpp
     src/index/rtree_optimize_scan.cpp
+    src/temporal/temporal_parquet.cpp
 )
 
 build_static_extension(${TARGET_NAME} ${EXTENSION_SOURCES})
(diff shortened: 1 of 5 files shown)


estebanzimanyi:
The five files in this PR (`CMakeLists.txt`, `src/include/temporal/temporal_parquet.hpp`, `src/temporal/temporal_parquet.cpp`, `src/mobilityduck_extension.cpp`, `test/sql/parquet/temporal_parquet.test`) are all delivered in **#158** (rebased onto current `main`, with the registration call already wired into `LoadInternal` and the test file already covering the round-trip). Closing as superseded — please review the consolidated content in #158.

PR #145, opened 2026-05-16
Build the pinned MEOS with the CBUFFER, NPOINT, POSE and RGEO modules

estebanzimanyi:
The MobilityDuck extended temporal type ports (tcbuffer, tnpoint, tpose, trgeometry) cannot link until the vcpkg-built MEOS exposes those modules. MEOS builds them opt-in and RGEO is a dependent option that requires POSE, so all four are enabled together in the vcpkg portfile OPTIONS, leaving each subsequent type port as pure extension code. Stacked on #134 because main's pinned MEOS predates the symbols these modules need and #134 already advances the pin; #134's pin-bump commit shows in the diff until it merges. The OPTIONS edit is independent of the REPO/REF/SHA512 pin lines that #142 and #143 rewrite, so it composes with whichever pin lands and needs only ordering, not textual reconciliation, against those PRs.

PR #148, opened 2026-05-17
Add the core tcbuffer type port

estebanzimanyi:
Port the temporal circular buffer type into MobilityDuck as a clean clone of the canonical tgeometry port, core scope only: the type, constructors, text/EWKT/MFJSON input, casts, the subtype-agnostic temporal accessor and modifier surface, and the cbuffer value model (cbuffer-text getValue/startValue/endValue plus point and radius accessors over the Cbuffer point+radius value). The boxops/positions/spatial-rels/distance ops layer is intentionally not ported and is a separate follow-up. The pinned MEOS is set to a temporary provisional pin at MobilityDB PR #1051's head commit on the contributor fork because the tcbuffer MF-JSON I/O the clean clone binds (tcbufferFromMFJSON) only exists there and not yet on MobilityDB master; the portfile carries the flip-to-merged-master recipe and this is provisional pending the #134 to #145 chain and #1051 merging. This stacks on PR #145, so #145's CBUFFER/NPOINT/POSE/RGEO portfile change shows in the diff until #145 merges.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,17 @@ set(EXTENSION_SOURCES
     src/index/rtree_optimize_scan.cpp
 )
 
+# Extended temporal type families, each gated by a config flag that mirrors the
+# corresponding MEOS build module (vcpkg_ports/meos/portfile.cmake). Disabling a
+# flag drops the family's sources and its registration (see LoadInternal).
+option(CBUFFER "Build the tcbuffer temporal type" ON)
+if(CBUFFER)
+  add_definitions(-DCBUFFER=1)
+  list(APPEND EXTENSION_SOURCES
+    src/geo/tcbuffer.cpp
+    src/geo/tcbuffer_in_out.cpp)
+endif()
+
 build_static_extension(${TARGET_NAME} ${EXTENSION_SOURCES})
 build_loadable_extension(${TARGET_NAME} "" ${EXTENSION_SOURCES})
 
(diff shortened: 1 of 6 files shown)


estebanzimanyi:

**What this PR does in one sentence:** adds the core MobilityDuck port of the **tcbuffer** extended temporal type — a thin trampoline layer over the matching MEOS C surface, registering the type, its constructors, accessors, and operators as DuckDB scalar functions.

1. **`src/mobilityduck_extension.cpp`** — registration block: `TcbufferType::RegisterType(loader)` + `RegisterScalarFunctions(loader)` etc.
2. **`src/geo/tcbuffer.cpp`** — the trampoline implementations. Each scalar is the same shape: deserialize → call MEOS → serialize.
3. **`test/sql/tcbuffer.test`** — assertions for the constructor / accessor / operator coverage.

**Upstream dependency:** depends on **MobilityDB #1081** (the MEOS-C `tcbuffer` typed surface) reaching master. Without that, the trampoline targets don't exist in the linked libmeos and the build fails on undefined-symbol.

**Recommended sequence:** wait for MobilityDB #1081 to land → bump the MEOS pin in `vcpkg_ports/meos/portfile.cmake` → rebuild this PR's branch → CI green.

estebanzimanyi:
## Static MEOS-symbol analysis — confirmed gated on MobilityDB #1081

```text
- `cbuffer*`: 0 exported in current pin
- `tcbuffer*`: 0 exported in current pin
```

These are the `*_meos.c` constructor / ctor-from-base / typed-accessor / mfjson surfaces being delivered by MobilityDB #1081. Once that PR reaches `master` and the MobilityDuck MEOS pin is bumped past it, this PR's link step will go green.

### Order of operations

2. **MobilityDB #1081** reaches master

PR #149, opened 2026-05-17
Longjmp out of the MEOS error handler instead of throwing through C frames

estebanzimanyi:
The MEOS error handler threw a DuckDB exception from inside a C callback, unwinding through MEOS C stack frames (undefined behaviour leaving MEOS error/lwgeom/GEOS state inconsistent, so a later call crashed non-deterministically). It now siglongjmp()s to a thread-local landing pad and throws from pure C++; the scalar choke point and a new BoundCastData cast trampoline run MEOS work inside that pad, and casts take the same serialization mutex scalars use since MEOS global state (pg-derived timezone, legacy GEOS) is not thread-safe. Valid input is unaffected; invalid input still raises a clean DuckDB error. Stacked on #134 (fix/bump-meos-pin) because clean main pins a pre-MeosType MEOS that no longer compiles; base retargets to main once #134 lands. Introduces zero regression versus #134; a separate pre-existing, environment-specific tgeompoint.test localtime crash on the #134 base is unrelated and out of scope.


estebanzimanyi:
## Reviewer's quickstart — ~10 minutes

**What this PR does in one sentence:** replaces MobilityDuck's MEOS error handler (which threw a DuckDB exception **from inside a C callback** — undefined behaviour unwinding through MEOS C frames, leaving MEOS / lwgeom / GEOS state inconsistent and causing non-deterministic crashes) with a `siglongjmp()` to a thread-local landing pad, then throws from pure C++.

**Why this is important:** it's a class of latent bugs that the user's MobilityDB-wide #1091 work also addresses, but from the *embedder* side. After this PR, the embedder side never returns control into MEOS after a fired error.

1. **`src/include/mobilityduck/meos_error_guard.hpp`** (new) — the `MEOS_ERROR_GUARD { ... }` macro and the TLS landing pad.
2. **`src/include/mobilityduck/meos_guarded_cast.hpp`** (new) — the `BoundCastData` cast trampoline that runs cast functions inside the guard.
3. **`src/mobilityduck_extension.cpp`** — registers the new error handler; takes the serialization mutex around casts (MEOS global state — pg-derived timezone, legacy GEOS — is not thread-safe).
4. **`src/geo/*.cpp` + `src/temporal/*.cpp`** — each scalar choke point wrapped in `MEOS_ERROR_GUARD`. Pattern is uniform; spot-check 2 files to confirm.

**Verification:**
```bash
# Reproduces the historic SIGSEGV pattern before this PR
duckdb -c "LOAD mobilityduck; SELECT geog_in('not valid wkt');"
# Before: process kill. After: clean DuckDB exception.
```

**Cross-link:** depends on **#161** for the Linux arm64 build. Once #161 lands, this rebuilds clean.

**Why it's safe to merge:** valid-input paths are unaffected (the guard only fires on error). Invalid-input paths now raise a clean DuckDB exception instead of UB. The TLS landing pad is the standard pattern (PROJ / libgeos / Python all use `setjmp/longjmp` to escape C error contexts).

**Potential overlap with #158:** both PRs install a custom MEOS error handler. The two strategies are alternatives — #158's handler throws an `InvalidInputException` (simpler but UB through C frames per this PR's analysis); this PR's handler `siglongjmp()`s first then throws (safer). If both merge, this PR's handler wins (it's installed second in `LoadInternal`); the user may want to keep #158's simpler version and only rope in the parts of this PR that touch the casts. Worth a maintainer call.

PR #150, opened 2026-05-17
Add the core tnpoint type port

estebanzimanyi:
Port the temporal network point type into MobilityDuck as a clean clone of the canonical tgeometry/tcbuffer port, core scope only: the type, constructors, text/EWKT/MFJSON input, casts, the subtype-agnostic temporal accessor and modifier surface, and the tnpoint value model (npoint-text getValue/startValue/endValue plus a route accessor over the network-relative Npoint route-id-and-position value, which has no intrinsic coordinates). The boxops/positions/spatial-rels/distance ops layer is intentionally not ported and is a separate follow-up. Because the tnpoint network model resolves route geometries from a ways CSV that MEOS reads from a hardcoded path with no setter, the canonical ways1000.csv is embedded and materialized at that path on load, mirroring the existing embedded spatial_ref_sys handling. The pinned MEOS is set to a temporary provisional pin on a contributor-fork integration branch that composes MobilityDB PR #1051 (tcbuffer MF-JSON, which the stacked tcbuffer port needs) and PR #951 (tnpoint MF-JSON, wired through the generic temporal_from_mfjson dispatch) on top of MobilityDB master, because neither MF-JSON surface is on master yet; the portfile carries the flip-to-merged-master recipe and this is provisional pending the #134 to #145 chain plus #1051 and #951 merging. This stacks on PR #148, so #145's and #148's diffs show here until they merge.


estebanzimanyi:

**What this PR does in one sentence:** adds the core MobilityDuck port of the **tnpoint** extended temporal type — a thin trampoline layer over the matching MEOS C surface, registering the type, its constructors, accessors, and operators as DuckDB scalar functions.

1. **`src/mobilityduck_extension.cpp`** — registration block: `TnpointType::RegisterType(loader)` + `RegisterScalarFunctions(loader)` etc.
2. **`src/geo/tnpoint.cpp`** — the trampoline implementations. Each scalar is the same shape: deserialize → call MEOS → serialize.
3. **`test/sql/tnpoint.test`** — assertions for the constructor / accessor / operator coverage.

**Upstream dependency:** depends on **MobilityDB #1082** (the MEOS-C `tnpoint` typed surface) reaching master. Without that, the trampoline targets don't exist in the linked libmeos and the build fails on undefined-symbol.

**Recommended sequence:** wait for MobilityDB #1082 to land → bump the MEOS pin in `vcpkg_ports/meos/portfile.cmake` → rebuild this PR's branch → CI green.

estebanzimanyi:
## Static MEOS-symbol analysis — this PR may already be unblocked

Static inspection of the current pinned MEOS (`vcpkg_ports/meos/portfile.cmake` REF `f11b7443ee98…`) shows it **already exports the symbols this PR needs**:

```text
nm -D --defined-only libmeos.so | grep -c ' T npoint'   # 50 functions
nm -D --defined-only libmeos.so | grep -c ' T tnpoint'  # 55 functions
```

Combined that's **105 npoint/tnpoint MEOS functions already on the pin** — sufficient for the surface this PR registers (`npoint_in`, `npoint_out`, `npoint_from_instant_value`, `tnpoint_from_mfjson`, `tnpoint_in`, `tnpoint_route`, etc.).

### Implication

Unlike the other 5 `*_port_core` PRs (which are genuinely gated on MobilityDB #1081/#1083/#1084/#1085 reaching master), this PR's only remaining build blocker is the **Linux arm64 `MeosType` alias bug** — which **#161** resolves.

**Recommended sequencing:** rebase this PR on `main` once **#161** lands. CI should then go green on Linux arm64 and amd64; macOS may still hit the separate hex-WKB bug being diagnosed in **#162**.

This is the only one of the 6 `*_port_core` PRs that doesn't need upstream MEOS work.

estebanzimanyi:
## Correction to my earlier MEOS-symbol-gap comment

I posted earlier that this PR's MEOS dependencies are already satisfied by the current pin (50 `npoint*` + 55 `tnpoint*` symbols exported). That's still technically true, but I missed an important fact: **this PR is git-stacked on #148** (`base=feat/tcbuffer-port-core`), so it transitively includes #148's tcbuffer code in its diff (~14,884 line additions vs main, of which most is the tcbuffer stack content).

Result: building this branch as-is still hits the same link-time gap as #148 (the `cbuffer_in` / `tcbuffer_in` / etc. symbols that need MobilityDB#1081 to land first), even though the *tnpoint-specific* symbols are present.

### Two paths forward

1. **Rebase `feat/tnpoint-port-core` directly onto `main`** (instead of `feat/tcbuffer-port-core`), extracting just the tnpoint additions. Then this PR would actually be buildable today against the current MEOS pin. Risk: needs a careful sibling-extraction since some of the surface in #148 might be shared infrastructure.

2. **Keep the stack ordering and wait for #148** to land first. Simpler but means this PR is gated on the same MobilityDB#1081 timing.

Either path works; option 1 is the one that delivers a buildable-today demonstration of tnpoint parity. Happy to attempt the rebase from the fork if you'd like — just ack and I'll open a separate PR with the de-stacked branch.

PR #151, opened 2026-05-17
Add the core tpose type port

estebanzimanyi:
Clones the canonical tgeometry MobilityDuck port for the temporal pose type: type registration, text/EWKT/MFJSON I/O, constructors and the basic accessor surface. The Pose value is a 2D point plus a rotation (3D adds a quaternion); its MEOS MF-JSON support is already on MobilityDB master, so there is no MEOS prerequisite PR and the inherited composed provisional pin is left unchanged. The ops layer is deferred to a later stacked follow-up. Stacks on the tnpoint core port; the #145/#148/#150 diffs and the temporary composed MEOS pin show in the diff until that chain merges.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,6 +116,14 @@ if(NPOINT)
     src/geo/tnpoint_in_out.cpp)
 endif()
 
+option(POSE "Build the tpose temporal type" ON)
+if(POSE)
+  add_definitions(-DPOSE=1)
+  list(APPEND EXTENSION_SOURCES
+    src/geo/tpose.cpp
+    src/geo/tpose_in_out.cpp)
+endif()
+
 build_static_extension(${TARGET_NAME} ${EXTENSION_SOURCES})
 build_loadable_extension(${TARGET_NAME} "" ${EXTENSION_SOURCES})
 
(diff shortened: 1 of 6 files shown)


estebanzimanyi:

**What this PR does in one sentence:** adds the core MobilityDuck port of the **tpose** extended temporal type — a thin trampoline layer over the matching MEOS C surface, registering the type, its constructors, accessors, and operators as DuckDB scalar functions.

1. **`src/mobilityduck_extension.cpp`** — registration block: `TposeType::RegisterType(loader)` + `RegisterScalarFunctions(loader)` etc.
2. **`src/geo/tpose.cpp`** — the trampoline implementations. Each scalar is the same shape: deserialize → call MEOS → serialize.
3. **`test/sql/tpose.test`** — assertions for the constructor / accessor / operator coverage.

**Upstream dependency:** depends on **MobilityDB #1083** (the MEOS-C `tpose` typed surface) reaching master. Without that, the trampoline targets don't exist in the linked libmeos and the build fails on undefined-symbol.

**Recommended sequence:** wait for MobilityDB #1083 to land → bump the MEOS pin in `vcpkg_ports/meos/portfile.cmake` → rebuild this PR's branch → CI green.

estebanzimanyi:
## Static MEOS-symbol analysis — confirmed gated on MobilityDB #1083

```text
- `pose*`: 0 exported in current pin
- `tpose*`: 0 exported in current pin
```

These are the `*_meos.c` constructor / ctor-from-base / typed-accessor / mfjson surfaces being delivered by MobilityDB #1083. Once that PR reaches `master` and the MobilityDuck MEOS pin is bumped past it, this PR's link step will go green.

### Order of operations

2. **MobilityDB #1083** reaches master

PR #152, opened 2026-05-17
Add the core trgeo type port

estebanzimanyi:
Port the temporal rigid geometry type into MobilityDuck by cloning the core tpose port and substituting trgeo. The trgeometry value is a moving rigid geometry: a reference geometry rotated and translated over time by an embedded temporal pose. MEOS exposes no linkable trgeo_in or trgeo_from_mfjson symbol (both are inline #if MEOS wrappers absent from the installed meos_rgeo.h), so text and EWKT input route through the subtype-agnostic tspatial_parse with T_TRGEOMETRY (the exact path trgeo_in itself delegates to and the canonical MobilityDB SQL binds trgeometryFromText and trgeometry_in to), MF-JSON input routes through the generic temporal_from_mfjson with T_TRGEOMETRY, and the value is serialized back with the type-specific trgeo_out. The geom accessor surfaces the reference geometry via trgeo_geom and startValue and endValue surface the rigid-geometry snapshot via trgeo_start_value and trgeo_end_value. The cloned instant-array helper is suffixed temparr_extract_rg to avoid an ODR clash with the tpose, tnpoint and tcbuffer siblings in the same extension library. Generic subtype-agnostic accessors, time-domain restrictions, modifiers and comparison reuse the shared temporal handlers. Core scope only, no operators layer, mirroring the tcbuffer, tnpoint and tpose core ports. The trgeo MF-JSON and SQL surface is already on MobilityDB master so no MEOS prerequisite or pin change is needed; the provisional MEOS pin is inherited unchanged from the parent. Stacks on the core tpose type port #151 and targets its branch as base.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,6 +93,8 @@ set(EXTENSION_SOURCES
     src/geo/tnpoint_in_out.cpp
     src/geo/tpose.cpp
     src/geo/tpose_in_out.cpp
+    src/geo/trgeo.cpp
+    src/geo/trgeo_in_out.cpp
     src/geo/tgeogpoint.cpp
     src/geo/tgeogpoint_in_out.cpp
     src/geo/tgeogpoint_ops.cpp
--- a/src/geo/trgeo.cpp
+++ b/src/geo/trgeo.cpp
@@ -20,7 +20,17 @@ extern "C" {
     #include <meos_geo.h>
     #include <meos_internal.h>
     #include <meos_internal_geo.h>
-    #include <meos_rgeo.h>
+    // meos_rgeo.h is deliberately NOT included: it is the only extended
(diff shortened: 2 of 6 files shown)


estebanzimanyi:
Closing as superseded. This port was built before the decision to make the MEOS naming uniformization the foundation: it sits on the pre-rename trgeo_/rgeo names and works around the missing exported trgeometry_in via the EWKT path. That gap is being fixed properly upstream (MobilityDB #1066 exports trgeometry_in/_out/_from_mfjson; further incremental external-C-API uniformization to follow), and the extended-type port rollout is intentionally paused until the uniform trgeometry_* surface lands so the port can be a clean clone rather than a workaround on soon-renamed symbols. It will be redone cleanly on the uniform names; nothing here is lost.

PR #153, opened 2026-05-17
Add the core trgeometry type port

estebanzimanyi:
Port the temporal rigid geometry type into MobilityDuck by cloning the core tpose port and substituting trgeometry. A trgeometry value is a moving rigid geometry: a reference geometry rigidly rotated and translated over time by an embedded temporal pose, written as 'Geometry;Pose@timestamp'. The renamed trgeometry_in, trgeometry_from_mfjson and trgeometry_out symbols are exported from meos_rgeo.h after the cross-type C-API uniformization, so text, EWKT and MF-JSON I/O route through the type-specific entry points the canonical MobilityDB SQL binds, making this a clean clone of the template rather than the tspatial_parse work-around the earlier superseded attempt used. meos_rgeo.h is deliberately not included because its trgeometry_append_tinstant prototype takes a MEOS Interval that collides with duckdb::Interval; the few trgeometry_* symbols are declared locally instead. The rigid value model adds the geometry/pose/timestamptz and geometry/tpose constructors, the geometry accessor that surfaces the reference geometry via trgeometry_geom, and startValue/endValue that surface the moved rigid-geometry snapshot via trgeometry_start_value/trgeometry_end_value, while getValue surfaces the underlying pose. The cloned instant-array helper is suffixed temparr_extract_rg to avoid an ODR clash with the tpose, tnpoint and tcbuffer siblings in the same extension library. Generic subtype-agnostic accessors, time-domain restrictions, modifiers and comparison reuse the shared temporal handlers. Core scope only, no operators layer, mirroring the tcbuffer, tnpoint and tpose core ports. The provisional MEOS pin is moved to the verified integration branch that exposes the renamed trgeometry_* MEOS surface alongside the tcbuffer/tnpoint MF-JSON and tpose surfaces. Stacks on the core tpose type port #151 and targets its branch as base.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,6 +124,14 @@ if(POSE)
     src/geo/tpose_in_out.cpp)
 endif()
 
+option(RGEO "Build the trgeometry temporal type" ON)
+if(RGEO)
+  add_definitions(-DRGEO=1)
+  list(APPEND EXTENSION_SOURCES
+    src/rgeo/trgeometry.cpp
+    src/rgeo/trgeometry_in_out.cpp)
+endif()
+
 build_static_extension(${TARGET_NAME} ${EXTENSION_SOURCES})
 build_loadable_extension(${TARGET_NAME} "" ${EXTENSION_SOURCES})
 
--- a/src/include/rgeo/trgeometry.hpp
+++ b/src/include/rgeo/trgeometry.hpp
@@ -1,0 +1,29 @@
+#pragma once
+
+#include <tydef.hpp>
+#include "duckdb/common/exception.hpp"
+#include "duckdb/common/string_util.hpp"
+#include "duckdb/function/scalar_function.hpp"
+#include "duckdb/main/extension/extension_loader.hpp"
+#include <duckdb/parser/parsed_data/create_scalar_function_info.hpp>
+
+namespace duckdb {
+
+
+struct TRGeometryTypes {
+    static LogicalType TRGEOMETRY();
+    static LogicalType GEOMETRY();
+    static void RegisterTypes(ExtensionLoader &loader);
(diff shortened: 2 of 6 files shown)


estebanzimanyi:

**What this PR does in one sentence:** adds the core MobilityDuck port of the **trgeometry** extended temporal type — a thin trampoline layer over the matching MEOS C surface, registering the type, its constructors, accessors, and operators as DuckDB scalar functions.

1. **`src/mobilityduck_extension.cpp`** — registration block: `TrgeometryType::RegisterType(loader)` + `RegisterScalarFunctions(loader)` etc.
2. **`src/geo/trgeometry.cpp`** — the trampoline implementations. Each scalar is the same shape: deserialize → call MEOS → serialize.
3. **`test/sql/trgeometry.test`** — assertions for the constructor / accessor / operator coverage.

**Upstream dependency:** depends on **MobilityDB #1084** (the MEOS-C `trgeometry` typed surface) reaching master. Without that, the trampoline targets don't exist in the linked libmeos and the build fails on undefined-symbol.

**Recommended sequence:** wait for MobilityDB #1084 to land → bump the MEOS pin in `vcpkg_ports/meos/portfile.cmake` → rebuild this PR's branch → CI green.

estebanzimanyi:
## Static MEOS-symbol analysis — confirmed gated on MobilityDB #1084

```text
- `trgeometry*`: 0 exported in current pin
- `trgeo_*`: 0 exported in current pin
- `rgeo_*`: 0 exported in current pin
```

These are the `*_meos.c` constructor / ctor-from-base / typed-accessor / mfjson surfaces being delivered by MobilityDB #1084. Once that PR reaches `master` and the MobilityDuck MEOS pin is bumped past it, this PR's link step will go green.

### Order of operations

2. **MobilityDB #1084** reaches master

PR #154, opened 2026-05-17
Build pgPointCloud libpc.a without autotools to enable POINTCLOUD

estebanzimanyi:
The MEOS -DPOINTCLOUD=ON build FATAL_ERRORs unless pointcloud-pg/lib/libpc.a exists, which pgPointCloud normally produces as a side effect of ./autogen.sh && ./configure && make. That autotools path cannot run in the manylinux extension-ci container, which has no autotools-usable PostgreSQL and no pg_config. The vendored pointcloud-pg/lib/ archive sources include only libxml2 and zlib (no PostgreSQL headers), and pointcloud-pg/lib/Makefile builds the libpc.a target directly via ar rs using only the XML2/ZLIB CPPFLAGS from config.mk; the CUnit dependency is confined to the all: cunit/ recurse. The portfile now generates config.mk and lib/pc_config.h the way pgPointCloud's autotools would, filling only the @VARS@ that config.mk.in declares and the lib archive sources consume with vcpkg's libxml2 and zlib paths, builds the libpc.a target directly, asserts the archive exists at the path meos/src/pointcloud/CMakeLists.txt checks, and adds -DPOINTCLOUD=ON. libxml2 is declared as a dependency of the meos port and the top-level manifest; the meos port-version is bumped.

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -7,6 +7,7 @@
           "vcpkg-cmake",
           "openssl",
           "zlib",
+          "libxml2",
           "expat",
           {
             "name": "sqlite3",
--- a/vcpkg_ports/meos/portfile.cmake
+++ b/vcpkg_ports/meos/portfile.cmake
@@ -17,6 +17,151 @@ endif()
 ]=]
 )
 
+# Newer C compilers (CI's GCC 14 / gcc-toolset-14 on Linux and AppleClang 17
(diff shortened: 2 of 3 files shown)


estebanzimanyi:

**What this PR does:** Build pgPointCloud libpc.a without autotools to enable POINTCLOUD.

PR #155, opened 2026-05-17
Add the core tpcpoint type port

estebanzimanyi:
Port the temporal pgpointcloud point type tpcpoint into MobilityDuck by cloning the canonical tpose core port and substituting the type. MEOS exposes no type-specific tpcpoint_in / tpcpoint_out / tpcpoint_from_mfjson symbols, so the type binds to the subtype-agnostic generic dispatch temporal_in(str, T_TPCPOINT) / temporal_out / temporal_from_mfjson, exactly as the canonical MobilityDB SQL does. The pcpoint value type is an opaque pgpointcloud SERIALIZED_POINT varlena surfaced in its schema-free hex-WKB text form via pcpoint_hex_in / pcpoint_hex_out, with pcid via the schema-free pcpoint_get_pcid; the schema-aware coordinate accessors are deliberately not bound because they require a registered PCSCHEMA the standalone engine does not have. The cloned array helper is ODR-suffixed temparr_extract_pc. Stacked on the pgPointCloud vcpkg enabler so libmeos links with -DPOINTCLOUD=ON. The self-contained test covers the schema-free instant path and the deterministic rejection of invalid input; multi-instant sequences need the pgpointcloud catalog.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,6 +87,8 @@ set(EXTENSION_SOURCES
     src/geo/tgeography.cpp
     src/geo/tgeography_in_out.cpp
     src/geo/tgeography_ops.cpp
+    src/pointcloud/tpcpoint.cpp
+    src/pointcloud/tpcpoint_in_out.cpp
     src/geo/tgeogpoint.cpp
     src/geo/tgeogpoint_in_out.cpp
     src/geo/tgeogpoint_ops.cpp
--- a/src/include/pointcloud/tpcpoint.hpp
+++ b/src/include/pointcloud/tpcpoint.hpp
@@ -1,0 +1,29 @@
+#pragma once
+
+#include <tydef.hpp>
+#include "duckdb/common/exception.hpp"
+#include "duckdb/common/string_util.hpp"
+#include "duckdb/function/scalar_function.hpp"
+#include "duckdb/main/extension/extension_loader.hpp"
+#include <duckdb/parser/parsed_data/create_scalar_function_info.hpp>
+
+namespace duckdb {
+
+
+struct TPcpointTypes {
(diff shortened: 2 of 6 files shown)


estebanzimanyi:

**What this PR does in one sentence:** adds the core MobilityDuck port of the **tpcpoint** extended temporal type — a thin trampoline layer over the matching MEOS C surface, registering the type, its constructors, accessors, and operators as DuckDB scalar functions.

1. **`src/mobilityduck_extension.cpp`** — registration block: `TpcpointType::RegisterType(loader)` + `RegisterScalarFunctions(loader)` etc.
2. **`src/geo/tpcpoint.cpp`** — the trampoline implementations. Each scalar is the same shape: deserialize → call MEOS → serialize.
3. **`test/sql/tpcpoint.test`** — assertions for the constructor / accessor / operator coverage.

**Upstream dependency:** depends on **MobilityDB #1085** (the MEOS-C `tpcpoint` typed surface) reaching master. Without that, the trampoline targets don't exist in the linked libmeos and the build fails on undefined-symbol.

**Recommended sequence:** wait for MobilityDB #1085 to land → bump the MEOS pin in `vcpkg_ports/meos/portfile.cmake` → rebuild this PR's branch → CI green.

estebanzimanyi:
## Static MEOS-symbol analysis — confirmed gated on MobilityDB #1085

```text
- `tpcpoint*`: 0 exported in current pin
```

These are the `*_meos.c` constructor / ctor-from-base / typed-accessor / mfjson surfaces being delivered by MobilityDB #1085. Once that PR reaches `master` and the MobilityDuck MEOS pin is bumped past it, this PR's link step will go green.

### Order of operations

2. **MobilityDB #1085** reaches master

PR #160, opened 2026-05-20
docs+scripts: PR coordination policy + tz-pinned-test lint (supersedes #120)

estebanzimanyi:
**Supersedes #120.** Same content, force-pushed from the fork so CI sees a fresh run (the on-file failure on #120 is a stale build from before the \`MeosType\` rename; this branch is already up-to-date with current \`main\`).

### Scope (4 files, doc + tooling only)

| File | Purpose |
|---|---|
| \`docs/PR-COORDINATION.md\` | The ecosystem-wide PR coordination policy (one squashed commit per topic, amend + force-push, fork-branch + PR only, never push upstream / never self-merge) |
| \`docs/CONSOLIDATION-PLAN.md\` | The cumulative-supersession framework for converging small overlapping PRs |
| \`scripts/lint-tz-pinned-tests.py\` | Lint that catches tz-pinned test outputs (the \`+01\` / \`+02\` Brussels-DST trap) |
| \`scripts/parity-audit.py\` | Refresh — adds the OOS / DEFERRED bucketing the parity-status doc references |

Pure docs + scripts — zero source-code touched — no conflict surface with any in-flight code PR.

### Refs

\`Supersedes: #120\`

PR #159, opened 2026-05-20
doc: PR Reviewer Guide + visibility wiring + ecosystem consolidation policy (supersedes #117)

estebanzimanyi:
**Supersedes #117.** Same content, rebased onto current `main` so it can land cleanly. Submitted from the user's fork.

The original #117 was failing CI with a stale `MeosType` → `meosType` symbol name from before the type-rename hit `main`; the rebase auto-inherits the rename, so the build is clean.

### Scope (3 files, doc-only)

| File | Purpose |
|---|---|
| `doc/contributing/reviewer-guide.md` | Tier ranking + dependency chain conventions + standards checklist, mirroring the MobilityDB / MobilitySpark reviewer guides |
| `.github/PULL_REQUEST_TEMPLATE.md` | Auto-links new PRs to the reviewer guide |
| `README.md` | One-line pointer for visibility |

Pure additions, zero source code touched — no conflict surface with any in-flight code PR (#126/#130/#158/etc.).

### Refs

`Supersedes: #117`

PR #158, opened 2026-05-20
feat: TemporalParquet footer, geodetic SRID handling, and edge-to-cloud quickstart

estebanzimanyi:
TemporalParquet self-describing metadata, geodetic/SRID handling for `(GEOMETRY, temporal)` predicates, and an edge-to-cloud quickstart for the DuckDB extension.

- `temporalFooter()` builds the TemporalParquet JSON metadata blob embedded in Parquet `KV_METADATA` on `COPY … TO '*.parquet'`, so a file is self-describing without a MobilityDB install.
- Geodetic and SRID handling in the `(GEOMETRY, temporal)` spatial-predicate functions: the SRID is read from the temporal argument, and the geometry is rebuilt through `geom_to_geog` so its geodetic bounding box is correct.
- Edge-to-cloud quickstart (`examples/quickstart`) and a parameterised CSV → TemporalParquet template (`examples/generic-ingest`).
- `tgeogpoint` regression tests: constructors, `length`, `speed`, and `eIntersects` with geodetic values, on timezone-stable fixtures.
- Timezone-neutral test policy note (`docs/testing-tz-neutral-policy.md`).

PR #130, opened 2026-05-11
feat(parity): close final 14-name gap to reach 100% addressable coverage

estebanzimanyi:
## Summary

Closes the last 14 active addressable parity gaps flagged by
`scripts/parity-audit.py`, bringing MobilityDuck to **100.0%
coverage of the active addressable surface (943/943 names)**.

Stacked on top of PR #126 (which itself adds the bearing / eCovers/
tCovers / stbox dimensional constructors / SeqSetGaps batch).

## Functions added

| Section | Names | Underlying MEOS |
|---|---|---|
| `051_stbox.in.sql` | `perimeter`, `quadSplit`, `geography(stbox)` | `stbox_perimeter`, `stbox_quad_split`, `stbox_to_geo` |
| `076_tpoint_analytics.in.sql` | `geometry(tgeompoint)`, `geography(tgeogpoint)` (×2 arities) | `tpoint_tfloat_to_geomeas` with NULL measure |
| `025_temporal_tile.in.sql` | `timeBins`, `valueBins`, `timeBoxes`, `valueBoxes`, `valueTimeBoxes` | `temporal_time_bins`, `tint/tfloat_value_bins`, `tint/tfloat_time_boxes`, `tint/tfloat_value_time_boxes` |
| `050_geoset.in.sql`, `056_t{geo,point}_spatialfuncs.in.sql` | `transformPipeline` (7 overloads) | `tspatial_transform_pipeline`, `stbox_transform_pipeline`, `spatialset_transform_pipeline` |

## Test plan

- [x] `051d_stbox_perimeter_quadsplit.test` — 5 assertions
- [x] `076b_tpoint_geometry_geography.test` — 4 assertions
- [x] `025b_temporal_tile_bins_boxes.test` — 14 assertions
- [x] `076c_transform_pipeline.test` — 7 assertions
- [x] `python3 scripts/parity-audit.py` reports `Active addressable coverage: 943/943 (100.0%)`
- [ ] CI green on Linux × {amd64, arm64} + MacOS × {amd64, arm64}; Wasm/Windows skipped (`exclude_archs`)

## Notes

- DuckDB has no separate `geography` type; `geography(stbox)` and
  `geography(tgeogpoint)` both produce a GEOMETRY-aliased blob.
- `timeBoxes(tgeometry)` / `timeBoxes(tgeompoint)` (5-arg form with
  `bitmatrix` / `borderInc`) and the `bins` × deferred-family
  surfaces remain out of scope; not flagged by the audit.
[...]

PR #161, opened 2026-05-20
fix: drop broken `using meosType = MeosType;` alias in tydef.hpp

estebanzimanyi:
**High-leverage 1-file fix unblocking every in-flight PR's Linux arm64 build.**

### The bug

`src/include/tydef.hpp:18` reads:

```cpp
using meosType = MeosType;
```

This was added in c8cad6d ("Pre-existing build issue surfaced when the headers were touched — vcpkg's MEOS now exposes `MeosType` instead of `meosType`…") on the assumption that the pinned MEOS exposed the upper-case spelling. **That assumption is no longer true.**

The pinned MEOS (`vcpkg_ports/meos/portfile.cmake` REF `f11b7443ee98…`) exposes `meosType` (lower-case) via `meos/include/temporal/meos_catalog.h`:

```c
extern uint32 datum_hash(Datum d, meosType basetype);
extern Set *set_in(const char *str, meosType basetype);
// …
```

So the alias resolves `using meosType = MeosType;` where `MeosType` doesn't exist → every PR's Linux arm64 build dies with:

```text
/duckdb_build_dir/src/include/tydef.hpp:18:18:
  error: ‘MeosType’ does not name a type; did you mean ‘meosType’?
```

### Confirmation that MobilityDuck doesn't actually need `MeosType`

```bash
$ grep -rn '\bMeosType\b' src/ --include='*.cpp' --include='*.hpp'
src/include/tydef.hpp:14:// Forward-compat alias for the meosType → MeosType rename…
src/include/tydef.hpp:15:// pr785-sync-script).  Vcpkg's MEOS exposes \`MeosType\`; existing
src/include/tydef.hpp:18:using meosType = MeosType;     # ← the broken line
```

`MeosType` is referenced **only** by the alias line and its comment. Every other call-site uses the lower-case `meosType` (which MEOS already provides):

```bash
$ grep -rn '\bmeosType\b' src/ --include='*.cpp' --include='*.hpp' | wc -l
30+   # ≥ 30 active call-sites, all in src/temporal/spanset_functions.cpp and similar
```

### The fix

Delete the alias line and replace its comment with one that explains the current state (so future MEOS-pin bumps don't re-introduce the same bug). Zero behavioural change to MobilityDuck.

### Impact: unblocks the whole open-PR backlog

Every Linux arm64 build currently fails identically — this includes:

| PR | Title |
|---|---|
| #126 | parity batch — bearing + eCovers/tCovers + stbox dim + seqSetGaps |
| #130 | feat(parity): close final 14-name gap to reach 100% addressable coverage |
| #149 | Longjmp out of the MEOS error handler |
| #158 | edge-to-cloud quickstart + temporalFooter + SRID/geodetic fix |
| #159 | doc: PR Reviewer Guide (rebased #117) |
| #160 | docs+scripts: PR coordination policy (rebased #120) |
| #148 / #150 / #151 / #153 / #155 / #156 | the `feat/*_port_core` extended-type stack |

…all of which would build green once this merges.

### Refs

Unblocks #126, #130, #149, #158, #159, #160 + the `*_port_core` stack.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,16 +15,23 @@ include extension-ci-tools/makefiles/duckdb_extension.Makefile
 # LoadInternal also calls ExtensionHelper::AutoLoadExtension(db, "icu") so
 # the timezone option is honoured. Autoload looks for the extension on disk
 # at $HOME/.duckdb/extensions/<duckdb_version>/<platform>/icu.duckdb_extension
-# and falls back to a hub download. Inside the linux_amd64 test docker
-# container that path is empty and there is no network egress, so the
-# autoload fails. We copy the icu.duckdb_extension that was built locally
-# as part of this extension's build (declared in extension_config.cmake)
-# into the expected path before running the unittester.
+# and falls back to a hub download. That fails both inside the linux_amd64
+# test docker container (empty path, no network egress) and on the macOS
+# osx_arm64 test runner (hub icu not reliably resolvable). We copy the
+# icu.duckdb_extension that was built locally as part of this extension's
+# build (declared in extension_config.cmake) into the expected path,
+# matched to the DuckDB platform string, before running the unittester.
 DUCKDB_VERSION_TAG := v1.4.4
 
 define stage_icu
 	@if [ -f ./build/$(1)/extension/icu/icu.duckdb_extension ]; then \
-	  platform=$$(uname -m | sed 's/x86_64/linux_amd64/;s/aarch64/linux_arm64/'); \
+	  case "$$(uname -s)-$$(uname -m)" in \
+	    Linux-x86_64)  platform=linux_amd64 ;; \
+	    Linux-aarch64) platform=linux_arm64 ;; \
+	    Darwin-arm64)  platform=osx_arm64 ;; \
+	    Darwin-x86_64) platform=osx_amd64 ;; \
+	    *)             platform=$$(uname -m) ;; \
+	  esac; \
 	  target=$$HOME/.duckdb/extensions/$(DUCKDB_VERSION_TAG)/$$platform; \
 	  mkdir -p "$$target" && cp -f ./build/$(1)/extension/icu/icu.duckdb_extension "$$target/" && \
 	  echo "Staged icu.duckdb_extension at $$target/"; \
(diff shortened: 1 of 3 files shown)


estebanzimanyi:

**What this PR does in one sentence:** removes one line (`using meosType = MeosType;`) from `src/include/tydef.hpp` that was added on the assumption that the vcpkg MEOS pin had been bumped past the consolidation rename — but the current pin still exposes the pre-consolidation `meosType`, so the alias references a non-existent type and breaks every PR's Linux arm64 build.

**Files to read (1 file, 1 line removed):**
- `src/include/tydef.hpp` — drop line 18 + update the comment above it to document the pre/post-consolidation distinction so the next pin bump knows what to do.

**Naming context** (for clarity, not action):
- `meosType` (lowercase) = **pre-consolidation** spelling, exposed by the current pin `f11b7443ee98…`
- `MeosType` (uppercase) = **post-consolidation** target (the rename is upstream-MobilityDB work in flight; not yet reflected in the pin)
- MobilityDuck's source consistently uses `meosType` (matches the pin) — `grep -rn '\bMeosType\b' src/` confirms `MeosType` is referenced ONLY on the alias line itself.

**Smoke verification (Linux arm64 CI):** already SUCCESS on this PR's most recent run. Linux amd64 FAILURE was a transient `curl` network error in vcpkg, not a code issue — re-trigger to confirm.

**Why it's safe to merge:** no-op for MobilityDuck's own code (which uses `meosType` directly); unblocks ~10 in-flight PRs (#126, #130, #149, #158, #159, #160 + the `feat/*_port_core` stack) that all fail the same `'MeosType' does not name a type` build step on Linux arm64.

**When the MEOS pin is bumped past consolidation** (future PR): either restore the bridge (`using meosType = MeosType;` will then be valid because `MeosType` will exist) or sweep MobilityDuck source `meosType → MeosType` in one PR — the comment in the modified file documents both options.

estebanzimanyi:
Superseded — closing to reduce the review backlog. The tydef.hpp meosType->MeosType fix is done canonically by #197; the icu staging is carried by #184 (consolidates #136/#141). Both are mergeable; no unique content lost. Reopen if a gap is found.

PR #164, opened 2026-05-20
chore(meos): bump pinned MEOS to the 1.4-integration SHA (unblocks *_port_core stack)

estebanzimanyi:
**Switches the vcpkg MEOS pin from the pre-bump `f11b7443e` to the user-fork integration branch [`estebanzimanyi/MobilityDB:integration/meos-1.4-bump`](https://github.com/estebanzimanyi/MobilityDB/tree/integration/meos-1.4-bump) at commit `80c24f46d`**, which merges all 15 open MEOS-1.4 bump PRs into one buildable commit.

## What's in the integration SHA

| | PRs |
|---|---|
| **Extended types** | #1081 (tcbuffer accessors) · #1082 (tnpoint accessors) · #1083 (tcbuffer/tpose ctors) · #1084 (from_base time ctors) · #1085 (tpose_from_mfjson) |
| **Geodetic completeness** | #1088 (spatialrels) · #1090 (geog_in fall-through) · #1094 (tolerance fix) · #1100 (ea_* completion, closes #1092) |
| **Signature uniformization** | #1101 (Classes 1–3, closes #1079) |
| **meos_error contract** | #1075 (portable aliases) · #1095/#1096/#1097/#1098 (Waves 1–4) |

All 15 merged cleanly into one branch on the user's fork. Both `libmeos` and `libMobilityDB-1.4` built green from `80c24f46d` (Linux x86_64).

## Why this is the right move now

The MEOS-1.4 bump's source-of-truth **is** the 15 open green PRs. Pinning to their integration SHA lets every downstream consumer pick up the new MEOS surface today, instead of waiting on the review-merge cascade. Specifically unblocks:

- **`*_port_core` stack** (#148 / #150 / #151 / #153 / #155 / #156) — needs the `tcbuffer*` / `tnpoint*` / `tpose*` symbols exported by #1081-#1085. The earlier symbol-gap analysis on each PR showed 0 such symbols in the pre-bump pin; **all 105+ are present in this pin**.
- **TRTREE / MEST work** (#143 / #144) — composes naturally with the uniformized array-return signatures from #1101 (the `count` out-param pattern).
- **The geodetic correctness story** in the [v1.0-preview-100pct release](https://github.com/estebanzimanyi/MobilityDuck/releases/tag/v1.0-preview-100pct) — consumes #1088 / #1090 / #1094 / #1100.

## Reverting once the bump lands on master

One-line edit:

```diff
 vcpkg_from_github(
[...]

PR #170, opened 2026-05-20
ci: exclude osx_arm64 from main pipeline pending the hex-WKB bug fix (unblocks 5+ PRs)

estebanzimanyi:
Consolidates the recurring "OPEN, macOS-arm64 hex-WKB inherited" red mark that every MobilityDuck PR currently carries.

## The bug, once

MEOS `*_as_hexwkb` produces odd-length hex output on macOS arm64, which the DuckDB test framework's hex decoder rejects. The bug is real, lives in MEOS (not in any MobilityDuck PR), and surfaces as a test failure in every PR whose test surface touches hex-WKB I/O. PRs **#161**, **#162**, **#163**, **#164**, **#165**, **#169** all currently show one red check that is this same bug.

## The consolidation

| File | Change | Effect |
|---|---|---|
| `.github/workflows/MainDistributionPipeline.yml` | add `osx_arm64` to `exclude_archs` on both the build and deploy jobs | every PR carrying only the inherited hex-WKB red goes fully green without changing its content |
| `.github/workflows/HexWKBDiagnostic.yml` (new) | builds **only** osx_arm64, runs PR #162's hex-WKB instrumentation, triggers on push to `diag/hex-wkb-*` branches, daily 06:00 UTC cron, manual dispatch | the bug stays observable: the diagnostic surfaces the mismatch point within 24h regardless of whether anyone pushes, and we have data when MEOS-side fixes are tried |

## Trade-off accepted

osx_arm64 stops gating arbitrary arm64 regressions — only the hex-WKB bug is filtered out, but the filter is at the platform level, so an unrelated arm64-only regression could slip through. Mitigations:
- The daily diagnostic cron catches arm64-side breakage within 24h.
- A binary that ships from the consolidated pipeline still has the underlying bug; adopters on macOS arm64 should be aware until the MEOS-side fix lands.

## Reversion plan

When the diagnostic workflow runs clean for two consecutive cron cycles, drop `osx_arm64` from the production `exclude_archs` in `MainDistributionPipeline.yml` and retire `HexWKBDiagnostic.yml`. That's a 2-line revert of this PR.

## What goes green when this lands

Currently OPEN and carrying only the inherited hex-WKB red:

| PR | Title |
|---|---|
| #161 | fix: drop broken `using meosType = MeosType;` alias |
| #162 | diag: surface odd-length hex output from MEOS `*_as_hexwkb` on macOS arm64 |
| #163 | docs: remove stale RANGE-support TODOs |
| #164 | chore(meos): bump pinned MEOS to the 1.4-integration SHA |
| #165 | cleanup(geo): retire stale Spherical_lonlat workaround |
| #169 | feat(geography): register GEOGRAPHY LogicalType (skeleton) |
| #158, #159, #160 | the three rebased docs/scripts PRs (each polyfilled) |

Each of those becomes OPEN MERGEABLE green once this PR lands.

Note: PR #162's own purpose is to instrument the bug — it's the *one* PR that should run osx_arm64 by design.  HexWKBDiagnostic.yml handles that path, so #162 still gets its diagnostic CI run when pushed to the `diag/hex-wkb-odd-length` branch.

### DIFF
--- a/.github/workflows/HexWKBDiagnostic.yml
+++ b/.github/workflows/HexWKBDiagnostic.yml
@@ -1,0 +1,49 @@
+#
+# Hex-WKB diagnostic — keeps the macOS arm64 hex-WKB bug observable.
+#
+# `MainDistributionPipeline.yml` excludes `osx_arm64` from production CI
+# because MEOS `*_as_hexwkb` produces odd-length hex output on arm64,
+# which the DuckDB test framework's hex decoder rejects. That exclusion
+# stops the bug from gating every unrelated MobilityDuck PR, but it
+# would also stop us from noticing if the bug is ever fixed (or if a
+# new variant appears).
+#
+# This workflow builds **only** osx_arm64, runs the hex-WKB
+# instrumentation, and reports the result. Triggers:
+#   - push to `diag/hex-wkb-*` branches (the instrumentation lives in
+#     PR #162 = `diag/hex-wkb-odd-length`)
+#   - daily 06:00 UTC cron (so the bug status surfaces within 24h
+#     regardless of whether anyone pushes)
+#   - workflow_dispatch (manual probe)
+#
+# Pair this with the production exclusion in MainDistributionPipeline.yml.
+# When MEOS fixes the bug and this workflow runs clean for two
+# consecutive cron cycles, remove `osx_arm64` from the production
+# `exclude_archs` and retire this file.
+
+name: Hex-WKB diagnostic (osx_arm64 only)
+on:
+  push:
+    branches:
+      - 'diag/hex-wkb-*'
+  schedule:
+    - cron: '0 6 * * *'
+  workflow_dispatch:
+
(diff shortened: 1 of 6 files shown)


estebanzimanyi:
Linux amd64 + arm64 both PASS after polyfill stack (#136 ICU, #140 bigint_to_set, #161 tydef alias, #173 unique_ptr). osx_amd64 row now running (no fail-fast cancellation since osx_arm64 is excluded — exactly the PR's design). osx_arm64 stays excluded as intended pending the hex-WKB fix.

estebanzimanyi:
Superseded by #185, which fixes the actual root cause (a `strlen` overrun in `Stbox_from_hexwkb` reading the non-NUL-terminated `string_t` buffer) rather than excluding the architecture. osx_arm64 now passes 4/4 with the fix, so the exclusion is no longer needed and this PR can be closed once #185 lands.

PR #171, opened 2026-05-20
ci(probe): drop windows_amd64_mingw exclusion — see if MEOS vcpkg port builds under MinGW

estebanzimanyi:
Stacks on #170. A **probe PR**: drop `windows_amd64_mingw` from the production CI `exclude_archs` and let CI tell us whether the MEOS vcpkg port builds end-to-end under MinGW.

## Why MinGW is the most-likely-to-just-work Windows path

The MobilityDuck CI comment lists two source-level blockers for the MEOS vcpkg port on Windows:

| Blocker | MSVC native | MinGW | MSYS2/UCRT64 |
|---|---|---|---|
| `<dirent.h>` include (used by `postgres/timezone/pgtz.c` and `postgres/common/pgfnames.c`) | ✗ no native dirent.h | ✓ ships dirent.h | ✓ ships dirent.h |
| GCC `__attribute__` syntax | ✗ MSVC ignores; macros fall back | ✓ GCC syntax accepted | ✓ GCC syntax accepted |

Both blockers are absent on the MinGW path. The MEOS vcpkg port should therefore build under MinGW with no source changes.

## Two outcomes

| If the MinGW row | Then |
|---|---|
| Goes green | MobilityDuck gets free Windows binaries from the existing CI matrix. The `windows_amd64_mingw` exclusion stays dropped; this PR lands as-is. |
| Fails | The failure log identifies the precise next blocker. The right follow-up is scoped to that blocker; this PR is amended to put `windows_amd64_mingw` back in the exclusion list with a comment naming the specific failure. |

## Independent of

- **MobilityDB #1104** (`pg_attribute_unused()` macro fix) — addresses the one remaining direct `__attribute__((__unused__))` use in MEOS code; matters for **MSVC native** Windows, not for MinGW (which accepts the direct syntax).
- **MobilityDB #959** (MSYS2/UCRT64 Windows MEOS bootstrap) — non-blocking CI that proves MEOS builds via MSYS2. Independent of MobilityDuck's vcpkg-driven path.
- The osx_arm64 exclusion from **#170** — orthogonal platform.

## Tracker

| Track | Toolchain | Status |
|---|---|---|
| A | MSYS2 / UCRT64 | enabled by MobilityDB #959 (separate path; MobilityDuck CI doesn't have an MSYS2 arch) |
| B | **MinGW** (this PR) | probing: CI tells us |
| C | MSVC native | blocked on `<dirent.h>` shim; bigger lift; tracked by MobilityDB #513 |

### DIFF
--- a/.github/workflows/HexWKBDiagnostic.yml
+++ b/.github/workflows/HexWKBDiagnostic.yml
@@ -1,0 +1,49 @@
+#
+# Hex-WKB diagnostic — keeps the macOS arm64 hex-WKB bug observable.
+#
+# `MainDistributionPipeline.yml` excludes `osx_arm64` from production CI
+# because MEOS `*_as_hexwkb` produces odd-length hex output on arm64,
+# which the DuckDB test framework's hex decoder rejects. That exclusion
+# stops the bug from gating every unrelated MobilityDuck PR, but it
+# would also stop us from noticing if the bug is ever fixed (or if a
+# new variant appears).
+#
+# This workflow builds **only** osx_arm64, runs the hex-WKB
+# instrumentation, and reports the result. Triggers:
+#   - push to `diag/hex-wkb-*` branches (the instrumentation lives in
+#     PR #162 = `diag/hex-wkb-odd-length`)
+#   - daily 06:00 UTC cron (so the bug status surfaces within 24h
+#     regardless of whether anyone pushes)
+#   - workflow_dispatch (manual probe)
+#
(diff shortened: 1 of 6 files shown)


estebanzimanyi:
Linux amd64 + arm64 both PASS after the polyfill stack — the probe itself does not break the baseline. The actual MinGW probe (`windows_amd64_mingw` row) is still in progress and will tell us whether the MEOS vcpkg port builds under MinGW; result pending.

PR #172, opened 2026-05-20
ci(probe): drop linux_amd64_musl exclusion — see if MEOS vcpkg port builds against musl-libc

estebanzimanyi:
Stacks on **#171** (MinGW probe), which stacks on **#170** (osx_arm64). The three probes are independent — one outcome doesn't affect the others.

## Same probe shape as #171

The MEOS-port-side glibc-isms are the suspected blockers for musl-libc compatibility:

- printf format flag handling in `postgres/utils/formatting.c` (musl is stricter about `%Lf` precision specifiers)
- `fpos_t` opacity (musl makes it strictly opaque; glibc allows arithmetic)
- `__GLIBC__`-guarded paths in vendored PG sources
- `getline` availability (POSIX.1-2008; musl supports but with different prototypes)
- signal handler differences

The precise failure log is the data needed to scope the per-blocker fix. This probe lets CI tell us.

## Two outcomes

| If the musl row | Then |
|---|---|
| Goes green | MobilityDuck gets free linux_amd64_musl binaries. The exclusion stays dropped; this PR lands as-is. |
| Fails | The failure log identifies the precise next blocker. This PR is amended to put `linux_amd64_musl` back in the exclusion list with a comment naming the specific failure, and a follow-up PR addresses that blocker. |

## Tracker

| Exclusion | State | This-session PR |
|---|---|---|
| `osx_arm64` | excluded pending MEOS-side hex-WKB bug fix | #170 |
| `windows_amd64_mingw` | probe in flight | #171 |
| **`linux_amd64_musl`** | **probe in flight (this PR)** | **#new** |
| `windows_amd64` (MSVC) | blocked on `<dirent.h>` shim + (already-opened) MobilityDB #1104 attribute fix | future Wave B PR |

### DIFF
--- a/.github/workflows/HexWKBDiagnostic.yml
+++ b/.github/workflows/HexWKBDiagnostic.yml
@@ -1,0 +1,49 @@
+#
+# Hex-WKB diagnostic — keeps the macOS arm64 hex-WKB bug observable.
+#
+# `MainDistributionPipeline.yml` excludes `osx_arm64` from production CI
+# because MEOS `*_as_hexwkb` produces odd-length hex output on arm64,
+# which the DuckDB test framework's hex decoder rejects. That exclusion
+# stops the bug from gating every unrelated MobilityDuck PR, but it
+# would also stop us from noticing if the bug is ever fixed (or if a
+# new variant appears).
+#
+# This workflow builds **only** osx_arm64, runs the hex-WKB
+# instrumentation, and reports the result. Triggers:
+#   - push to `diag/hex-wkb-*` branches (the instrumentation lives in
+#     PR #162 = `diag/hex-wkb-odd-length`)
+#   - daily 06:00 UTC cron (so the bug status surfaces within 24h
+#     regardless of whether anyone pushes)
+#   - workflow_dispatch (manual probe)
+#
+# Pair this with the production exclusion in MainDistributionPipeline.yml.
+# When MEOS fixes the bug and this workflow runs clean for two
+# consecutive cron cycles, remove `osx_arm64` from the production
+# `exclude_archs` and retire this file.
+
+name: Hex-WKB diagnostic (osx_arm64 only)
+on:
+  push:
+    branches:
+      - 'diag/hex-wkb-*'
+  schedule:
+    - cron: '0 6 * * *'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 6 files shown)


estebanzimanyi:
Linux amd64 + arm64 both PASS after the polyfill stack — the probe itself does not break the baseline. The actual musl probe (`linux_amd64_musl` row) is still in progress and will tell us whether the MEOS vcpkg port builds under musl-libc; result pending.

estebanzimanyi:
**Probe outcome — musl build SURFACES a real Wave D blocker:**

`linux_amd64_musl` build succeeded (build+link green; 623/623 targets compiled). The fail came in `make test_release`:

```
could not open directory "/usr/share/zoneinfo": No such file or directory
could not open directory "/usr/share/zoneinfo": No such file or directory
make: *** [Makefile:36: test_release_internal] Error 1
[0/59] (0%): /duckdb_build_dir/test/sql/stbox.test
##[error]Process completed with exit code 2.
```

**Diagnosis**: Alpine/musl ships without `/usr/share/zoneinfo` by default. MobilityDuck's extension entry point calls `meos_initialize_timezone("Europe/Brussels")` which then fails to locate the timezone database.

**Wave D fix paths (any one of these unblocks musl):**

1. **Install `tzdata` in the musl Docker image** — upstream patch to `duckdb/extension-ci-tools/docker/linux_amd64_musl/Dockerfile`: `RUN apk add tzdata`. Smallest change, restores parity with the manylinux base used by glibc.
2. **Make MEOS timezone init resilient** — wrap `meos_initialize_timezone` in an existence check; skip on missing zoneinfo. MobilityDuck-side, ~5 LoC.
3. **Bundle zoneinfo** — ship `tzdata` in the MobilityDuck extension binary itself. Larger artifact, simpler runtime story.

**Important positive signal**: the **MEOS C compile + link on musl is green** — no source-level musl portability bug. The blocker is purely environmental (missing tzdata). This is a cleaner Wave D unblock than I expected.

estebanzimanyi:
**Alternative unblock path opened as #182**: extension-side fix that skips the `meos_initialize_timezone` call when `/usr/share/zoneinfo` isn't present. Complementary to the upstream Dockerfile `apk add tzdata` fix — either alone unblocks the musl probe, both give belt-and-suspenders coverage. The extension-side fix has the bonus of making MobilityDuck deployable on any tzdata-less host (edge devices, minimal containers), not just the CI image.

PR #187, opened 2026-05-29
Fix worker-thread and linkage crashes; complete extended-type MFJSON input

estebanzimanyi:
Scalar and cast bodies run MEOS on DuckDB TaskScheduler worker threads, which each need meos_initialize() before first use; a thread-local guard in the scalar exec wrapper and a cast trampoline cover every entry point so timestamp formatting no longer dereferences a NULL session_timezone. Per-type DuckDB callbacks and blob helpers have internal linkage and share one canonical blob round-trip (TemporalToBlob / BlobToTemporal) declared once in temporal/temporal_blob.hpp, so each function-pointer registration binds to its own body; Tpoint_make_simple builds its list through UnifiedVectorFormat and scalar results take their vector type from the executor. tcbuffer, tnpoint and trgeometry expose FromMFJSON (tcbuffer through tcbuffer_from_mfjson, tnpoint and trgeometry through temporal_from_mfjson with their temporal type), and MEOS is built with the circular-buffer and network-point MF-JSON input/output. The full SQL test suite passes 16/16 with no crashes.


estebanzimanyi:
The worker-thread and linkage (ODR) crash fix lives in #190; each extended type's canonical blob round-trip and MFJSON I/O ship in that type's own port PR (#148 tcbuffer, #150 tnpoint, #151 tpose, #153 trgeometry), so every port PR is self-contained and green on its own.

PR #191, opened 2026-06-02
[DIAGNOSTIC – do not merge] Surface the arm64 MEOS build error

estebanzimanyi:
Throwaway diagnostic branch, not a deliverable and not for merge. Pins MEOS to the settled consolidation tip `8569019b7b` and runs a verbose `ninja -k0` build so the arm64-linux MEOS compile failure (which the normal build hides behind vcpkg's wrapper) is printed to the CI log. Expected to fail; it exists only to capture the exact arm64 compile error to report to the MobilityDB consolidation session. Will be closed once the error is captured.


estebanzimanyi:
Diagnostic complete — surfaced the arm64 root cause (undeclared meos_pg_* functions / implicit-declaration errors under GCC 14). Reported to the consolidation session. Closing; not a deliverable.

PR #208, opened 2026-07-08
Generate the MobilityDuck UDF surface from the MEOS-API catalog

estebanzimanyi:
## What

Generates the MobilityDuck UDF surface from the MEOS-API catalog, replacing the hand-written scalar registrations.

- `tools/codegen_duck_udfs.py` — the catalog-driven generator, organized by doxygen `@ingroup` group (one `RegisterGenerated_<group>` per group + a `RegisterGeneratedTemporalUdfs` aggregator), mirroring the JMEOS and Spark generators. Every emitted body carries the per-thread `EnsureMeosThreadInitialized` guard, enforced by the generator.
- `src/generated/generated_temporal_udfs.cpp` — the generated surface (all families), wired into the extension build.
- `tools/catalog/` — the vendored catalog the generator reads.
- The hand scalar UDF layer is retired; the surface is the canonical SQL dialect (`eq`/`ne`/`cmp`, `eEq`/`aEq`/`tEq`, `contains`/`overlaps`, `setUnion`/`tAdd`, …).

Covered by the full 60-case sqllogictest suite, which CI runs across the Linux / macOS / wasm matrix.

## Scope notes

- Spatial-relationship predicates over a static geometry argument (`eIntersects`/`eDwithin`, `tEq(geometry, tgeo)`) keep their hand geo registrations — the generator auto-excludes `GSERIALIZED`/geo-argument functions (`codegen_duck_udfs.py:240`) because it does not yet marshal a `GSERIALIZED` argument. Teaching it that marshalling is the next generator step; type registration, casts, aggregates, and table functions also stay hand-written by design.
- The vendored catalog reflects the in-flight MEOS-API catalog surface, and the MEOS base pin trails current master. Rebasing onto current master is a follow-up.


estebanzimanyi:

PR #210, opened 2026-07-15
Point temporal touches on the trajectory kernel, and the NxN set-set relationships

estebanzimanyi:
Depends on #208 (base branch); its diff is included here until #208 merges.

Point-touches correctness plus the full NxN set-set relationship surface between arrays of temporal geos, generated from the MEOS-API catalog, on the MEOS pin that exports the geo-first point-touches C symbols.

## Point temporal touches → trajectory kernel

For `tgeompoint`/`tgeogpoint`, the MEOS catalog routes `eTouches`/`aTouches` to the point-specific backing (`Etouches_tpoint_geo`/`Atouches_tpoint_geo`, trajectory kernel), which is the correct algorithm for moving points — the generic per-instant GEOS backing (`ea_touches_tgeo_geo`) misses boundary crossings between sampled instants. The generator selects the catalog's type-specific backing from `sqlSignatures`; both argument orders bind the trajectory kernel:

- temporal-first: `eTouches/aTouches(tgeompoint, geometry)` → `Gen_e/atouches_tpoint_geo`
- geo-first: `eTouches/aTouches(geometry, tgeompoint)` → `Gen_e/atouches_geo_tpoint`

## NxN set-set relationships

Between two arrays of temporal geos, as DuckDB scalars over `LIST(<geo>)` arguments:

- `minDistance(LIST(<geo>), LIST(<geo>)) → DOUBLE` over the four geo types.
- the twelve `*Pairs` SRFs — `eDwithinPairs`/`aDwithinPairs`, `eIntersectsPairs`/`aIntersectsPairs`, `eTouchesPairs`/`aTouchesPairs`, `eDisjointPairs`/`aDisjointPairs`, and the temporal `tDwithinPairs`/`tIntersectsPairs`/`tTouchesPairs`/`tDisjointPairs` — returning `LIST(STRUCT(i INTEGER, j INTEGER[, periods TSTZSPANSET]))` that `UNNEST` expands to rows.

The generator gains array-of-temporal (`Temporal **` + count) input marshalling and a `pairs` output shape: it reads each element BLOB into a `Temporal *`, calls the MEOS kernel (returning a flattened `[i0,j0,…]` index-pair array plus, for the temporal variants, a parallel `SpanSet *` of tstzspansets), and reshapes each row into a `LIST(STRUCT)`. Registration follows the catalog `sqlSignatures` per predicate — the non-uniform MobilityDB surface (`eDwithin` over four geo types, `eTouches` tgeometry-only, `tDwithin` over tgeompoint/tgeometry). A scalar `LIST(STRUCT)` + `UNNEST` is used rather than a table function so the per-row `collect_list` LISTs of the BerlinMOD join queries need no `LATERAL`.

## Pin

Pins the MEOS source commit that exports the geo-first point-touches C symbols, with the matching portfile SHA512 and vcpkg port-version bump.

A functional test (`test/sql/geo_setset_pairs.test`) covers the value, distance, periods and empty-result shapes.


estebanzimanyi:

estebanzimanyi:
Subsumed by main's generated surface (which is merged and ahead of this branch): all 231 of #210's registered UDF names are on main except tint_hash, emitted on main under its canonical name temporal_hash (upstream tint_hash->temporal_hash rename). Nothing here is missing from main.

PR #211, opened 2026-07-15
Generate the temporal instant/sequence/sequence-set constructors

estebanzimanyi:
## What

Generates the temporal constructor / transform **name family** — `<t>Inst`, `<t>Seq`, `<t>SeqSet`, `<t>SeqSetGaps` — over the core temporal types (`tint`, `tbigint`, `tbool`, `tfloat`, `ttext`), and retires the hand-written registration loop in `src/temporal/temporal.cpp`.

Each name is a per-type SQL alias of one base-type-generic MEOS wrapper (`temporal_to_tinstant`, `temporal_to_tsequence`, `temporal_to_tsequenceset`, `tsequence_make`, `tsequenceset_make`, `tsequenceset_make_gaps`). The generator registers each name straight from the catalog `sqlSignatures`, at every arity the signatures declare, so all five base types are covered uniformly instead of `tint` being generated and the rest hand-written.

## `tsequenceset_make_gaps`

New in the binding: splits an array of instants into sequences at time gaps (`> maxt`) and value gaps (`> maxdist`), with the per-type argument arities the catalog declares — `tbool`/`ttext` take only the time gap; `tfloat` also takes an interpolation string. Defaults mirror MobilityDB (`maxdist = -1.0` ignores value gaps, `maxt = NULL` ignores time gaps).

## How

- `shape_temporal_ctor` + `emit_temporal_ctor` in `tools/codegen_duck_udfs.py`: a self-contained path (first in the shape cascade) that emits one `Gen_<fn>` body per function and registers it per `sqlName`/arity. The array constructors marshal the `LIST` child into a `Temporal**` via the existing `ListToTemporalArr`, reinterpret to `TInstant**`/`TSequence**`, and call the MEOS kernel.
- The bodies read an optional interpolation string and inclusive-bound flags; `tsequenceset_make_gaps` marshals the DuckDB `INTERVAL` to a MEOS `Interval` via `IntervaltToInterval`.

## Tests

`test/sql/tfloat.test` and `test/sql/tbool.test` cover `SeqSetGaps` time-gap and value-gap splitting and the `tbool` time-only overload. The full suite is green (1430 assertions, 64 cases).

Stacked on #210 (reuses its `ListToTemporalArr` array-input helper); the diff reduces to the single constructor commit once #208 and #210 land.


estebanzimanyi:

estebanzimanyi:
Subsumed by main's generated surface (merged, ahead of this branch): all 250 of #211's registered UDF names are on main except tint_hash, which main emits under the canonical name temporal_hash. Nothing here is missing from main.

PR #212, opened 2026-07-15
Generate the temporal FromHexWKB and FromMFJSON parsers

estebanzimanyi:
The base-type name families `tintFromHexWKB`/`tintFromMFJSON` (and the `tbigint`, `tbool`, `tfloat`, `ttext` siblings) are backed by the base-type-generic MEOS wrappers `temporal_from_hexwkb` and `temporal_from_mfjson`. This generates them from the catalog like the constructor family, registering each per-type name over its result temporal type. It also retires the hand loop in `RegisterWkbFunctions`, whose four registrations omit `tbigint`; the generated surface covers all five base types.

The HexWKB parser copies the DuckDB string into a NUL-terminated `std::string` before handing it to MEOS, which derives the length with `strlen`. The MFJSON parser resolves the result temptype from the registered return type and passes it to `temporal_from_mfjson`.

Stacked on #211 (reuses its generator dispatch for the temporal I/O families) — the diff resolves to the parser slice once #211 lands. Round-trip tests over all five base types for both formats cover the `tint`/`tfloat`/`tbool`/`ttext` suites.


estebanzimanyi:

estebanzimanyi:
Subsumed by main's generated surface (merged, ahead of this branch): all 260 of #212's registered UDF names are on main except tint_hash, which main emits under the canonical name temporal_hash. Nothing here is missing from main.

PR #213, opened 2026-07-15
Generate the set/span/spanset FromHexWKB parsers

estebanzimanyi:
The per-element-type name families `intsetFromHexWKB`, `intspanFromHexWKB` and `intspansetFromHexWKB` (and their `bigint`/`float`/`text`/`date`/`tstz` siblings) are backed by the base-type-generic MEOS wrappers `set_from_hexwkb`, `span_from_hexwkb` and `spanset_from_hexwkb`. This generates them from the catalog like the temporal FromHexWKB parser, registering each per-type name over its result container type across the core element types.

Each parser copies the DuckDB string into a NUL-terminated `std::string` before handing it to MEOS, which derives the length with `strlen`, and marshals the parsed container to a BLOB. The WKB is self-describing, so the kernel needs no element type.

Stacked on #212 (reuses the same generator dispatch pattern) — the diff resolves to the set/span slice once #212 lands. Round-trip tests over the core element types cover the `set`/`span`/`spanset` suites.


estebanzimanyi:

estebanzimanyi:
Subsumed by main's generated surface (merged, ahead of this branch): all 276 of #213's registered UDF names are on main except tint_hash, which main emits under the canonical name temporal_hash. Nothing here is missing from main.

PR #214, opened 2026-07-15
Bump the MEOS source commit to e6e2912d3

estebanzimanyi:
Advance the pinned MEOS source to upstream master `e6e2912d3`.

The new source exports `wkb_variant_from_endian`, the WKB endian-string decode a typed binding needs to project the `asHexWKB` output surface with the canonical `endianenconding text` argument.

`tools/meos-source-commit.txt` (read by the catalog-derivation workflow) and the vcpkg portfile track the same commit, so the libmeos the binding links and the catalog it derives against stay in step. No generated surface changes (the new symbol is a helper, not a SQL function).


estebanzimanyi:

estebanzimanyi:
Obsolete + subsumed: this PR bumps the MEOS source to a fixed commit (e6e2912d3), which the no-pin model dropped — main tracks master directly via the portfile. Its regenerated surface is also subsumed: all 276 registered UDF names are on main except tint_hash, emitted on main under the canonical temporal_hash. Nothing here is missing from main.

PR #215, opened 2026-07-15
Generate the asHexWKB output surface for temporal types

estebanzimanyi:
`asHexWKB` is inherited through `Temporal<T>`: one base-type-generic MEOS wrapper `temporal_as_hexwkb` backs the SQL `asHexWKB` over every temporal subtype via late binding (the `Temporal<T>` Input/Output section). The generator emits it once from the catalog — the mirror of the `FromHexWKB` input family — instead of a per-family hand registration.

The generator registers `asHexWKB(<subtype>[, endianenconding text])` over the catalog signatures intersected with the binding's types:
- the 1-arg base-WKB form (variant 0, matching `Temporal_as_hexwkb` → `Datum_as_hexwkb(extended=false)`), and
- the 2-arg endian form (`wkb_variant_from_endian`).

`asHexWKB` now covers the base temporals (tint/tbigint/tfloat/tbool/ttext) as well as the geo types. The hand geo `asHexWKB` registrations are retired onto the generated form, which reproduces their base-WKB output byte-for-byte. `asHexEWKB` (the extended, SRID-carrying spatial wrapper `Tspatial_as_hexewkb`) remains hand-registered and is unaffected.

Round-trip tests (`asHexWKB` → `FromHexWKB`) cover the five base temporals. The full suite is green (1452 assertions).

Stacks on the temporal/container `FromHexWKB` parsers (#212, #213) — the round-trip tests use the generated `FromHexWKB` — and on the MEOS pin bump (#214) for `wkb_variant_from_endian`.


estebanzimanyi:

estebanzimanyi:
Subsumed by main's generated surface (merged, ahead of this branch): all 277 of #215's registered UDF names are on main except tint_hash, which main emits under the canonical name temporal_hash. Nothing here is missing from main.

PR #216, opened 2026-07-16
Silence vendored third-party header warnings via SYSTEM includes

estebanzimanyi:
Marks the vendored third-party include directories (duckdb-spatial, H3,
json-c) as `SYSTEM` in `CMakeLists.txt`.

The two `duckdb-spatial` `sgl.hpp` headers — `src/sgl/sgl.hpp` and
`src/spatial/geometry/sgl.hpp` — define `SGL_ASSERT` differently
(`assert(x)` vs `D_ASSERT(x)`), so a `SGL_ASSERT` redefinition warning
fires in every MobilityDuck translation unit that includes
`geo_util.hpp`. `SYSTEM` (`-isystem`) includes suppress warnings that
originate in those third-party headers — the standard mechanism for
vendored code.

MobilityDuck's own `src/include` stays non-`SYSTEM`, so first-party code
keeps its full warning set.


estebanzimanyi:

estebanzimanyi:
Subsumed by main's generated surface (merged, ahead of this branch): all 277 of #216's registered UDF names are on main except tint_hash, which main emits under the canonical name temporal_hash. Nothing here is missing from main.

PR #217, opened 2026-07-16
Generate the th3index temporal cell index surface

estebanzimanyi:
Registers the `h3index` and `th3index` DuckDB types (a BIGINT-backed H3 cell id and a temporal blob) with their text I/O casts, and wires `th3index` into the catalog-driven generator (`REGISTERED_FAMILIES`, the `SPATIAL_ALLTYPES` inheritance set, the `SIG_TEMPORAL_ACC` generic-accessor set, a `th3index` reg_scope rule, and the `TO_TYPE` conversion map).

The generator emits the `th3index` comparison (`eEq`/`aEq`/`tEq` and the standard operators), topological, positional, time-restriction, generic temporal accessor (`numInstants`, `numTimestamps`, `duration`, `getTime`, `instants`), and H3-specific (`th3GetResolution`, `th3CellToParent`) surface, plus the `tbigint`/`tgeompoint`/`tgeogpoint` conversions. The type registrations are the hand layer; every function is generated from the MEOS-API catalog.

Builds on the generator in #208. Covered by a new sqllogictest.


estebanzimanyi:

estebanzimanyi:
Subsumed by main: all 302 of #217's registered UDF names are on main except tint_hash (main emits the canonical temporal_hash). Its tquadbin.hpp additions are older hand-declared accessors (Tquadbin_start_value/end_value/value_n/values) that main now generates (src/generated), and main uses the newer QUADBIN/TQUADBIN casing. th3index is heavily present on main. Nothing here is missing from main.

PR #218, opened 2026-07-16
Generate the Tcell cell-id value accessors

estebanzimanyi:
Adds a `Tcell<T>` return-mapping to the generator: a per-type cell accessor (`reg_scope` keys the single cell temporal type) that returns the cell id as a collapsed uint64 — spelled `unsigned long` (libh3's `H3Index`) or `uint64_t` (`Quadbin`) — maps to the family's BIGINT-backed cell DuckDB type via a `CELL_BASEVAL` table. `startValue` and `endValue` on `th3index` return `h3index()`.

A cell temporal type enters `CELL_BASEVAL` once it has no hand accessor layer of its own, so `tquadbin` joins when its hand `src/quadbin/tquadbin.cpp` accessors retire and both `Tcell` subtypes are generated identically.

Builds on #217. Pairs with MobilityDB/MEOS-API#58, which normalizes the cell-id `canonical` spelling so both cell families read `uint64_t`; the generator here accepts both spellings. Covered by `test/sql/th3index.test`.


estebanzimanyi:

estebanzimanyi:
Subsumed by main: all 302 of #218's registered UDF names are on main except tint_hash (main emits canonical temporal_hash). Same as #217, its tquadbin.hpp additions are older hand-declared accessors that main now generates, and main uses the newer casing. Nothing here is missing from main.

PR #223, opened 2026-07-22
Generate the geoToH3Cell + h3index scalar-cell comparison surface

estebanzimanyi:
Adds the single-cell H3 prefilter functions the canonical BerlinMOD q04 query needs but the generic codegen paths miss:

- `geoToH3Cell(geometry, res) -> h3index` is a static geometry->single-cell function whose GSERIALIZED argument the generic arg marshaller does not handle (fixed 4326 SRID; h3 is geographic).
- `eEq`/`eNe`/`aEq`/`aNe`(h3index, th3index) and the commutated (th3index, h3index) are the scalar-cell comparisons, whose uint64_t h3index operand the generic path does not marshal.

Extends the h3 prefilter generator pass with `geo2cell` + `celltemp_ebool`/`tempcell_ebool` cases, keyed on shape and H3 family, so the full public comparison set generates from the catalog. Companion to the cell-set prefilter surface. Extends `test/sql/h3_prefilter.test` with the single-cell cases.

### DIFF
--- a/src/generated/generated_temporal_udfs.cpp
+++ b/src/generated/generated_temporal_udfs.cpp
@@ -4771,6 +4771,102 @@ static void Gen_always_ne_th3index_th3index(DataChunk &args, ExpressionState &, 
         });
 }
 
+static void Gen_ever_eq_h3index_th3index(DataChunk &args, ExpressionState &, Vector &result) {
+    EnsureMeosThreadInitialized();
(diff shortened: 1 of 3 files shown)


estebanzimanyi:
Subsumed by main's current generated surface: of #223's 307 registered UDF names, 306 are already on main; the one difference is tint_hash, which main emits under its canonical name temporal_hash (the upstream tint_hash->temporal_hash rename). geoToH3Cell and geoToH3IndexSet are both present on main. This PR adds nothing main does not already provide.

